PR #35, opened 2015-01-05
Set several default initial args to `None` for convenience

### LINKED ISSUE #29: Make sure that arguments that can take None as a value have a default value
A couple of methods recently got changed so that they can accept `None` for certain arguments that were previously marked `not None` (e.g. `acquire_cred`).  These methods should have those arguments have default values of `None` so that the methods don't have to be called like `acquire_cred(None)`.

frozencemetery:
This should be all occurrences in the low-level api, and so this should close #29.

More defaults can be set, but:

1. That would change the API since default arguments follow positionals, so this could not be for 1.1.0.
2. The convenience gained in these changes appears minimal at the present time.
3. This introduces further deviation in the low-level API from the C API, which makes it probably irrelevant.

### DIFF
--- a/gssapi/raw/creds.pyx
+++ b/gssapi/raw/creds.pyx
@@ -80,7 +80,7 @@ cdef class Creds:
             self.raw_creds = NULL
 
 
-def acquire_cred(Name name, lifetime=None, mechs=None, usage='both'):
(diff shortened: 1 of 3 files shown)


DirectXMan12:
LGTM :+1:, just remove the 'may need to make `not None`' comment from above 'acquire_cred_from'

PR #37, opened 2015-01-06
Set hostname in travis build

frozencemetery:
Setting the hostname should remove errors when attempting to establish a security context.

Additionally, robustify the installation of components from sssd in case we encounter version mismatch problems in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ python:
   - "3.4"
 
 install:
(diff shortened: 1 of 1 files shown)


DirectXMan12:
modulo krb5-multidev, LGTM :+1:

frozencemetery:
krb5-multidev is installed when libkrb5-dev is requested, so we get it regardless.

PR #39, opened 2015-01-20
Use GSS.framework on Mac OS X

sigmaris:
On Mac OS X, Apple wants us to use the GSS.framework to compile against GSSAPI, and `#include <GSS/...>` instead of `<gssapi/gssapi.h>`. With this change we can choose headers to link against at compile time, without changing our source and header files.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,3 +2,4 @@ include *.txt
 recursive-include docs *.txt
 recursive-include gssapi *.pxd
 recursive-include gssapi *.c
+recursive-include gssapi *.h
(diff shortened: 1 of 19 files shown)


DirectXMan12:
In general, this looks good.  Just a couple of small inline comments.

frozencemetery:
I concur, this looks good to me as well.  I think that creating header files to do this indirection is absolutely the right way to go.  That said, this version does not work.

In file "ext_cred_imp_exp.pyx", you need to change over to "python_gssapi_ext.h" from "gssapi/gssapi_ext.h", I think.

Barring that @DirectXMan12's changes, this looks good to me.

frozencemetery:
With new changes this is working, and looks good to me, though it needs to be squashed before ending up in the main repo.

PR #40, opened 2015-01-22
_with_password functions from Solaris/IAKERB

frozencemetery:
This should close #5.  As noted in the files themselves, gss_add_cred_with_password cannot currently be tested on anything other than MIT Kerberos master.

PR #41, opened 2015-01-22
Honor path when retrieving K5Test utilities

DirectXMan12:
This commit makes k5test.py honor the setting $PATH when calling
the various utilities it uses to set up its environment
(technically it uses `which`, so if your `which` binary does something
else, you may have a problem).


frozencemetery:
This tests good for me using kerb master installed into /opt/krb5 (for testing #40).

PR #42, opened 2015-02-03
Add krb5 version 1.13 as an additional build target

frozencemetery:
Krb5 1.13 is the latest released version, so we should be building against it.  This uses my ppa built against precise, which is itself a reworking of the Debian packaging.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,16 @@ python:
   - "3.3"
   - "3.4"
 
+env:
+  - KRB5_PPAS="sssd/updates"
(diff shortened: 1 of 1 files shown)


frozencemetery:
Ah, sorry that this is unclear.  My repository depends on things packaged in the sssd/updates repository (I didn't want to build things like libverto again, since they already have), so in both cases you need the sssd/updates repo, but you only need/want mine if you're doing 1.13 things.  Does this look better?

DirectXMan12:
LGTM :+1:

PR #43, opened 2015-02-10
Low-level: allow mutating input creds in add_cred

### LINKED ISSUE #18: Support mutating the existing credentials object with add_cred
We should support mutating the current credentials object with `add_cred`, which is possible if `NULL` is passed as the output cred object in the C API.  We should follow up with a change to the high-level API

DirectXMan12:
This commit adds an optional parameter to `add_cred`,
`mutate_input`, which allows `add_cred` to either
create a new set of credentials containing the input
credentials and the new credentials (False) or
to just modify the input credentials, adding the new
credentials directly to them (True).  Previously, only
the former was possible.

Closes #18.


frozencemetery:
Awesome, LGTM as long as travis agrees

PR #52, opened 2015-02-24
Make the method of extending enums a bit more sane

DirectXMan12:
This new way doesn't involve nearly as much hackery.
However, enum extensions must be in a separate submodule
(a module under `gssapi.raw._enum_extensions`), and should
be called the same thing as the corresponding extension
(due to the way that `setup.py` is written).  Additionally,
the import of the aforementioned submodules *must* happen
before importing any other python-gssapi modules (hence they
have to happen at the top of 'gssapi/raw/__init__.py').  This
limits the ability to allow for third parties writing extension
bindings (although that's probably not something that should be
done).  Finally, due to the way it works, each extenable enum
must be decorated beforehand, and must not share a name with
any other extendable enum.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@
 dist
 gssapi/**/*.c
 docs/build
+__dont_use_cython__.txt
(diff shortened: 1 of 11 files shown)


frozencemetery:
This seems superior to, and cleaner than, the current approach.  Looks good to me.

PR #53, opened 2015-02-25
Automatically upload checksumed taball on release

DirectXMan12:
This uploads a compressed tarball, as well as a checksum, for every tagged release.
The files will be uploaded once the Python 3.4 run of the krb5 1.13 (the latest released
version) tests completes.  The tarball has essentially the same contents as the github tarball.
However, since it is only created once, it has a consistent checksum (unlike the GitHub tarball,
which is created on demand, and thus may have different checksums).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ dist
 gssapi/**/*.c
 docs/build
 __dont_use_cython__.txt
+**/__pycache__
+.eggs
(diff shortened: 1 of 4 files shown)


frozencemetery:
It seems like it would be cleaner to remove the `$KRB5_PPAS` variable entirely and just use `$KRB_VER`, modifying the install step (probably moving it to a script) to install PPAs based on that.

Otherwise, this seems good to me, once I figure out why we're doing it.

PR #54, opened 2015-03-05
RFC 6680 Support

DirectXMan12:
This introduces both low-level and high-level support for RFC 6680.
In order for the for the attribute methods to run, the `greet_client` example
authdata plugin must be installed.  Otherwise, the attribute tests will be automatically
skipped.

Further explanation for how the low- and high-level APIs work may be found in
the commit messages.


frozencemetery:
By itself, this looks good to me.  More work needs to be done getting the greet_client integrated into our test process (I expect this is my problem), and the 1.10 stuff needs to be resolved as well before we can merge, I think.

frozencemetery:
Okay, I like the way the 1.10 fix looks.  This seems good to me.

PR #55, opened 2015-03-12
Actually use acceptor creds when passed in

DirectXMan12:
Previously, the condition on acceptor credentials was inverted,
such that when `acceptor_creds` was `None`, it would attempt to use
the value of the underlying raw `gss_cred_id_t`, and when it was
not `None`, the actual credentials passed in would be
`GSS_C_NO_CREDENTIAL`.  This has been corrected.

Additionally, two new tests have been added.  One checks that passing
None for acceptor credentials is ok, and the second properly tests that
`accept_sec_context` received S4U2Proxy delegated credentials in the
appropriate circumstances.

### DIFF
--- a/gssapi/raw/sec_contexts.pyx
+++ b/gssapi/raw/sec_contexts.pyx
@@ -297,7 +297,7 @@ def accept_sec_context(input_token not None, Creds acceptor_creds=None,
         output_context = SecurityContext()
 
     cdef gss_cred_id_t act_acceptor_cred
-    if acceptor_creds is None:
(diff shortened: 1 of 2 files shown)


frozencemetery:
LGTM; will merge once it passes travis.

simo5:
LGTM to me too

DirectXMan12:
I've disabled one of the tests on 1.10 since there seems to be a bug in the 1.10 s4u2proxy code that prevents the s4u delegated credentials from working properly.

PR #56, opened 2015-03-13
krb-console utility

DirectXMan12:
This commit introduces support for the krb-console utility, which is an interactive Python console which automatically sets up and tears down
a self-contained krb5 realm.  This allows the user to experiment around with the GSSAPI bindings without having to worry about
having or changing an existing krb5 setup.

This commit also introduces a utility method for locating the plugins dir, which was necessary to fix the k5test configuration
when run wholly under a LD_LIBRARY_PATH to a newer installation.

### DIFF
--- a/gssapi-console.py
+++ b/gssapi-console.py
@@ -1,0 +1,142 @@
+#!/usr/bin/env python
+
+# interactive console with easy access to krb5 test harness stuff
+
+import code
+import os
+import sys
+import copy
+
+from gssapi.tests import k5test
+
+
+READLINE_SRC = """
+# python startup file
+import readline
+import rlcompleter
+import atexit
+import os
+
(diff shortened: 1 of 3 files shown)


simo5:
Given this console can be quite generic, can we call it gss-console and give it options like --mech=krb5 to set up different environments ?
At some point I'd like to add --mech=ntlmssp and have it set up gss-ntlmssp as the mechanism and appropriate configuration for it.

DirectXMan12:
@simo5 : sounds like a good plan.  I'll update the code.

frozencemetery:
The plugin detection is tested working on Debian, which was the source of the problem, and includes and override mechanism for non-system installs as requested.

The console utility is likewise good to have, and looks good to me.

PR #59, opened 2015-03-19
Fix Setup Bugs

DirectXMan12:
This PR fixes the "multiple -DHAS_GSSAPI_EXT_H" bug and the bug that can cause custom libraries to not be linked against, even when specified by krb5-config.


frozencemetery:
This looks good to me.  Non-standard install locations are working with the second commit.  Will merge once it's past Travis.

PR #60, opened 2015-03-20
Fix plugin detection for subdirectories of /usr/lib

### LINKED ISSUE #57: Package greet_client for consumption in travis suite
We want to graft greet_client onto the repo(s) we're using for later versions of Kerberos.  In theory, since there has been no change to the greet_client code since before the 1.10 release, it is not a problem to only test it there and we don't need to worry about 1.12 and 1.10.

This was originally blocking #54 though it has been decided to continue without explicit greet_client testing in Travis.

frozencemetery:
Combined with changes to the krb5-1.13 packaging, this will cause the
greet_client tests to pass on travis.

Closes #57.


DirectXMan12:
:+1:

PR #61, opened 2015-04-09
use gssapi_ext wrapper to fix build with heimdal

pefoley2:


frozencemetery:
Hi, thanks for the patch!  This looks good to me, and does indeed fix the build with Heimdal.

PR #65, opened 2015-04-23
Automatically Upload to PyPI

DirectXMan12:
This commit automatically uploads releases to PyPI under the same
conditions that would trigger an upload of a tarball/checksum to
GitHub releases (docs are uploaded too).


frozencemetery:
One comment inline.  Otherwise, seems reasonable.

frozencemetery:
:+1:, pending Travis.

PR #70, opened 2015-06-05
Don't inquire about incomplete security contexts

### LINKED ISSUE #69: MIT krb5 refuses to inquire about partially established contexts
MIT krb5 refuses to inquire about partially established contexts, meaning `SecurityContext.complete` will fail unless it would return `True`.  This hasn't been an issue so far, since MIT krb5 negotiation seems to generally complete in one round trip.  However, it could be an issue in the future.

DirectXMan12:
This commit changes `SecurityContext` to handle a couple of cases where
we called `_inquire` when the context was not fully established.  A bug
in MIT krb5 prevents this from working properly.

In the case of `complete`, we simply cache the result of the last
`step` call.

Fixes #69

### DIFF
--- a/gssapi/sec_contexts.py
+++ b/gssapi/sec_contexts.py
@@ -110,10 +110,21 @@ class SecurityContext(rsec_contexts.SecurityContext):
 
         else:
             # we already have a context in progress, just inspect it
-            if self.locally_initiated:
-                self.usage = 'initiate'
-            else:
-                self.usage = 'accept'
+            # NB(directxman12): MIT krb5 refuses to inquire about a context
+            # if it's partially established, so we have to check here
+
+            try:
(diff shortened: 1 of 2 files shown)


frozencemetery:
Any chance you could specify what versions of MIT are affected (e.g., "MIT Kerberos up to version X.YY is affected...")?

When the fix is landed in MIT, we'll probably want to note it as well.

DirectXMan12:
> Any chance you could specify what versions of MIT are affected (e.g., "MIT Kerberos up to version X.YY is affected...")?

I figured I'd do that there's a version in which it does work

frozencemetery:
Open pull request: https://github.com/krb5/krb5/pull/285

PR #73, opened 2015-07-28
Fix error message locale behavior when LC_MESSAGES is set

### LINKED ISSUE #72: LookupError hiding GSSError when LC_MESSAGES is set
gssapi/raw/misc.pyx:291 - error handling incorrectly sets encoding to language name instead of encoding name (selecting the first element of tuple instead of second) when LC_MESSAGES is set, causing LookupError to be raised, hiding the original exception

frozencemetery:
Fixes #72 (issue is that we were using the wrong component of the tuple returned by getlocale).  This would only show up when LC_MESSAGES is set, which in most cases it is not.

### DIFF
--- a/gssapi/raw/misc.pyx
+++ b/gssapi/raw/misc.pyx
@@ -288,7 +288,7 @@ class GSSError(Exception, metaclass=GSSErrorRegistry):
                 given code
         """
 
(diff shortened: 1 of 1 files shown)


DirectXMan12:
LGTM :+1:

PR #74, opened 2015-08-10
Fix SPNEGO crash

frenche:
Hello,

This fix a segfault when trying to init_sec_context with SPNEGO mechanism.

Issue was found at (see there for an example to reproduce):
https://github.com/modauthgssapi/mod_auth_gssapi/pull/49

Thanks!

### DIFF
--- a/gssapi/raw/sec_contexts.pyx
+++ b/gssapi/raw/sec_contexts.pyx
@@ -204,7 +204,7 @@ flags=None, lifetime=None, channel_bindings=None, input_token=None)
         input_token_buffer.value = input_token
         input_token_buffer.length = len(input_token)
 
-    cdef gss_OID actual_mech_type
+    cdef gss_OID actual_mech_type = GSS_C_NO_OID;
     cdef gss_buffer_desc output_token_buffer = gss_buffer_desc(0, NULL)
     cdef OM_uint32 ret_flags
     cdef OM_uint32 output_ttl
@@ -232,7 +232,9 @@ flags=None, lifetime=None, channel_bindings=None, input_token=None)
 
     cdef OID output_mech_type = OID()
     if maj_stat == GSS_S_COMPLETE or maj_stat == GSS_S_CONTINUE_NEEDED:
-        output_mech_type.raw_oid = actual_mech_type[0]
+        if actual_mech_type:
+            output_mech_type.raw_oid = actual_mech_type[0]
+
         return InitSecContextResult(output_context, output_mech_type,
                                     IntEnumFlagSet(RequirementFlag, ret_flags),
                                     output_token,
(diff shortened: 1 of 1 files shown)


simo5:
Thanks!

gssapi/raw/sec_contexts.pyx
```diff
@@ -204,7 +204,7 @@ flags=None, lifetime=None, channel_bindings=None, input_token=None)
         input_token_buffer.value = input_token
         input_token_buffer.length = len(input_token)
 
-    cdef gss_OID actual_mech_type
+    cdef gss_OID actual_mech_type = GSS_C_NO_OID;
```
DirectXMan12: please remove the `;` from the end of this line

gssapi/raw/sec_contexts.pyx
```diff
@@ -232,7 +232,9 @@ flags=None, lifetime=None, channel_bindings=None, input_token=None)
 
     cdef OID output_mech_type = OID()
     if maj_stat == GSS_S_COMPLETE or maj_stat == GSS_S_CONTINUE_NEEDED:
-        output_mech_type.raw_oid = actual_mech_type[0]
+        if actual_mech_type:
```
DirectXMan12: you might want to make this `is not GSS_C_NO_OID` or `!= GSS_C_NO_OID` (your code works fine because `GSS_C_NO_OID` is actually `NULL`, but it's a bit clearer)

DirectXMan12:
a couple of nits inline.  Travis had a moment before, but it looks like you're just failing due to flake8 not liking the `;`

frenche:
Hi Solly and Simo,

On Mon, Aug 10, 2015 at 11:00 PM, Solly <notifications@github.com> wrote:

> a couple of nits inline. Travis had a moment before, but it looks like
> you're just failing due to flake8 not liking the ;
>

Thanks, I'll fix and add a commit.

One more thing, what do you think - do we want to raise an error if the
context has completed and we got no actual mech or we should trust the
library?

Regards,
Isaac B.

frozencemetery:
Isaac Boukris <notifications@github.com> writes:

> One more thing, what do you think - do we want to raise an error if the
> context has completed and we got no actual mech or we should trust the
> library?

I think we want to follow the behavior of the library, since it's either
a bug in the library (which in general we shouldn't workaround if
possible) or permitted by specification.

frenche:
On Mon, Aug 10, 2015 at 11:36 PM, Robbie Harwood <notifications@github.com>
wrote:

> Isaac Boukris <notifications@github.com> writes:
>
> > One more thing, what do you think - do we want to raise an error if the
> > context has completed and we got no actual mech or we should trust the
> > library?
>
> I think we want to follow the behavior of the library, since it's either
> a bug in the library (which in general we shouldn't workaround if
> possible) or permitted by specification.


Make sense - thanks.

frozencemetery:
Pushed to master as https://github.com/pythongssapi/python-gssapi/commit/89197da687f7ccc3847e3fc54c26ea395111e27c and https://github.com/pythongssapi/python-gssapi/commit/102ea36a8a0bc3375ab4b2dcc3f4fa4bd9b6ded3

PR #84, opened 2015-09-25
Don't require enum34 in python >= 3.4

frozencemetery:
This fixes the build in distros which no longer provide python3-enum34.

I'm really sorry to be the one introducing conditionals here after fighting against it elsewhere, but there does not seem to be another way to do this.


DirectXMan12:
LGTM :+1: once You fix the flake8 issues.

frozencemetery:
This is related to [this thing over here](https://github.com/pythongssapi/debian-python-gssapi/blob/174bd7e7d2a51377e78eeb3f2457088b9f256027/debian/patches/enum34.patch)

Issue #86, opened 2015-11-18
Specify decorator and six in basic requirements


frozencemetery:
Currently, python-gssapi needs python-six and python-decorator available to work.  We need to actually specify this.

PR #88, opened 2015-11-19
Correct requirements list

frozencemetery:
Includes decorator and six in the README, and also updates setup.py to ensure
we have a good version of six.

Closes: #86


DirectXMan12:
LGTM :+1:

You might want to also add "After being sure to install the requirements," in the "From the Git Repo" section.

PR #98, opened 2016-02-24
[travis] Switch to using Trusty builder

frozencemetery:
Going forward, it looks like Travis will be pushing migration to Trusty, which is reasonable.  This will also need to happen if we want to migrate to containers as far as I can tell.  More information on what this entails in [Travis's writeup](https://docs.travis-ci.com/user/trusty-ci-environment/).


frozencemetery:
Their documentation is incorrect - `apt-get` still prompts unless the forcing flags are passed.

frozencemetery:
(We have opted for the approach in #109 instead.)

Issue #96, opened 2016-02-24
the code for accepting delegated credentials is incorrect


simo5:
Both the low level and the high level API unconditionally try at any step of the context extablishment (on the acceptor side) to return delegated credentials and create the appropriate classes.

Delegated credentials are returned only if the context establishment return GSS_S_COMPLETE and are undefined in any other step.

In the high level API care need to be taken so that if the lower level returns None the delegated_creds should be set straight to None as well, as calling Credentials(None) will istead try to fetch the default credentials for the process, and that is an error.

DirectXMan12:
Recording conversations for posterity:

RFC 2744 says:  

> Only valid if deleg_flag in ret_flags is true, in which case an explicit credential handle (i.e. not GSS_C_NO_CREDENTIAL) will be returned; if deleg_flag is false, gss_accept_context() will set this parameter to GSS_C_NO_CREDENTIAL.

RFC 2743 says (in Inquire_context):

> GSS_S_COMPLETE indicates that the referenced context is valid and that deleg_state, mutual_state, replay_det_state, sequence_state, anon_state, trans_state, prot_ready_state, conf_avail, integ_avail, locally_initiated, and open return values describe the corresponding characteristics of the context.

And says for init_sec_context (also applying to accept_sec_context):

> These state indicators' _[deleg_state, etc]_ values are undefined unless either the routine's major_status indicates GSS_S_COMPLETE, or TRUE prot_ready_state is returned along with GSS_S_CONTINUE_NEEDED major_status; for the latter case, it is possible that additional features, not confirmed or indicated along with TRUE prot_ready_state, will be confirmed and indicated when GSS_S_COMPLETE is subsequently returned.

So it looks like we should check for `COMPLETE or (prot_ready and CONTINUE)`, and then check `DELEG_FLAG` (the equivalent of `deleg_state` in 2744).

DirectXMan12:
So the question is: does "undefined" mean "up to the implementation" or "in an invalid state" here?
If it means "up to the implementation", it seems to me according to 2744, we can count on:

- if DELEG_FLAG is false, delegated_creds is NULL
- if DELEG_FLAG is true, delegated_creds is *valid*
- DELEG_FLAG will always be either true or false

meaning that it's fine to just check NULL.  My concern is that if we add more complicated logic than "is it NULL or not" and miss a valid pointer, we'll have a memory leak, since 

[...]

Issue #95, opened 2016-02-23
`SecurityContext#delegated_creds` should be a property with documentation


DirectXMan12:
Currently, `SecurityContext#delegated_creds` is just a normal instance attribute of `SecurityContext`, which means that it can be accidentally overwritten, and it does not show up in the documentation.  This should be fixed.

simo5:
I think it should be possible to drop the delegated credentials esp if we allow to pickle the whole security context and then restore it.
Someone may want to pass around a security context but not transfer any delegated credentials to a lower priviliged process as dellegated credentials may include a delegated tgt.

DirectXMan12:
Pickling the `SecurityContext` uses a special pickle process that doesn't actually pickle the object -- it just uses `export_sec_context()` (so the unpickling process just reconstitutes the `SecurityContext` using `import_sec_context()`).

simo5:
This means delegated credentials get lost in the process.
On the one hand I think this is more secure, OTOH it may come as a suprise to the suer and should be carefully documented.

DirectXMan12:
Yeah.  The only reason delegated creds are stored as a field is so that the "step" method can return only a token.  It should definitely be documented, though

PR #99, opened 2016-02-25
Fix Delegated Credentials Issues

DirectXMan12:
This PR contains a fix for #95 (exposing `delegated_creds` as a property).

It also contains a partial fix for #96 (setting `delegated_creds` to `None` if no delegated credentials are returned by the low-level API, and only wrapping the result in a high-level `Credentials` API if delegated credentials are actually returned), which should have while we debate a longer-term solution.

### DIFF
--- a/gssapi/sec_contexts.py
+++ b/gssapi/sec_contexts.py
@@ -511,7 +511,10 @@ class SecurityContext(rsec_contexts.SecurityContext):
         res = rsec_contexts.accept_sec_context(token, self._creds,
                                                self, self._channel_bindings)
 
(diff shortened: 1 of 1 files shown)


DirectXMan12:
cc @simo @frozencemetery 

I'm thinking we can cut 1.1.5 after b8c76a0, and cut 1.2.0 after 45c9ee1 (since exposing a previously undocumented field would probably count as a backwards-compatible feature introduction)

frozencemetery:
Okay by me to merge once it's past travis.

PR #104, opened 2016-03-29
Force overwrite on cred_store tests

frozencemetery:
Certain versions of krb5 (>= 1.14 or anything that backported the
patch, including RHEL/Centos/Fedora) object (probably rightly so) to
duplicate credentials already existing in the ccache.  Tell those
versions to sit down and shut up.

This is the third and final commit in a series to enable running our
test suite on RHEL-7.

For posterity, to see the rest of the series, look at my nomerge/rhel7_patched branch.  The first commit was merged as pythongssapi/k5test#1 and the second (a workaround for old cython versions) is not planned for merge at this time.


DirectXMan12:
LGTM :+1:

PR #105, opened 2016-05-13
Issue template

frozencemetery:
I've been threatening this a while, and here it is.  I tried to keep it as simple as possible, in line with what I believe are @DirectXMan12's wishes.  There is no pull request template because, quite frankly, I don't know what I'd put in that; our PRs from external contributors have not been missing information that we need, and I haven't had a problem with that in other projects.

Let the bikeshedding begin!

### DIFF
--- a/ISSUE_TEMPLATE
+++ b/ISSUE_TEMPLATE
@@ -1,0 +1,13 @@
+### What went wrong?
+
+
+### How do we reproduce?
+
+*(Remember to use fenced code blocks and consider placing in a gist if large)*
+
+
+### Component versions (python-gssapi, Kerberos, OS / distro, etc.)
+
+*(Please include MIT/Heimdal/etc. and how you installed python-gssapi)*
+
+


ISSUE_TEMPLATE
```diff
@@ -0,0 +1,8 @@
+### What went wrong?
+
+
+### How do we reproduce?
```
DirectXMan12: You might want to add something like `(remember to use fenced code blocks when including code in the issue)` and/or `(if it involves a large code snippet, please place it in a gist)` (I don't think we've really seen either of those issues here, but I've seen them elsewhere, and it's nice avoid them preemptively).

ISSUE_TEMPLATE
```diff
@@ -0,0 +1,8 @@
...
+
+
+### Component versions (python-gssapi, Kerberos, OS, etc.)
```
DirectXMan12: `Kerberos distribution (MIT, Heimdal, etc) and version`, perhaps (so that we don't have to guess Kerberos distro by version number formatting)?

DirectXMan12:
Generally looks good.  A couple of suggestions inline.

frozencemetery:
Okay.  What do you think of this?

DirectXMan12:
LGTM :+1:

PR #109, opened 2016-11-24
Test unbreaking travis through docker

frozencemetery:
Travis will not pass until https://github.com/pythongssapi/k5test/pull/3 is merged, and k5test passes.  Currently only the RPM builders are broken on the latest k5test release.


.travis.sh
```diff
@@ -1,4 +1,53 @@
...
+        # if we don't have it, gcc doesn't work
+        yum install -y redhat-rpm-config
+    fi
+
+    yum -y install krb5-{devel,libs,server,workstation} which gcc findutils
+
+    if [ x"$PYTHON" = "x3" ]; then
+        yum -y install python3-pip python3-devel
+
+    # really not happy about this
```
DirectXMan12: salty comments are salty :-P

.travis.sh
```diff
@@ -20,7 +69,7 @@ F8_SETUP=$?
 flake8 gssapi
 F8_PY=$?
 
-flake8 gssapi --filename='*.pyx,*.pxd' --ignore=E225,E226,E227,E901,E402
+flake8 gssapi --filename='*.pyx,*.pxd' --ignore=E225,E226,E227,E901,E402,E999
```
DirectXMan12: we should probably note what each of these are that we're skipping, and why.

DirectXMan12:
If we do go with this, we need to make sure it doesn't break the release mechanisms at all.

DirectXMan12:
Looks like this adds between 30s and 1min to each test in the matrix, which is a bit of a time increase.  It's not horrible, though.  I wonder if we can ship images pre-installed with our test environment to speed things up?

DirectXMan12:
I do really like that we can now test across multiple distros easily, though.

frozencemetery:
(Addressed two comments in new version.)

It's more time in the matrix, but it's far less work for me since I don't have to get each version of krb5 (and to a lesser extent, Python) working on Ubuntu-14.04.  It also allows us to be more flexible with the underlying Travis infrastructure: we don't care what Ubuntu (or Linux, even) version underlies it.  It even brings us closer to not needing sudo, if I understand it right.

As for pre-installed images, I don't know; you know much more about container building than I do.

frozencemetery:
Sorry, against 12.04; this build transparently upgrades to 14.04.

Issue #111, opened 2017-02-15
display_status wrappers go cyclic


tiran:
### What went wrong?
GSSAPI failed with ```RuntimeError: maximum recursion depth exceeded in cmp``` while reporting an error from ```gssapi.raw.ext_cred_store.acquire_cred_from```

```
[Wed Feb 15 12:35:46.414055 2017] mod_wsgi (pid=2647): Exception occurred processing WSGI script '/usr/share/ipa/wsgi.py'.
[Wed Feb 15 12:35:46.414087 2017] Traceback (most recent call last):
[Wed Feb 15 12:35:46.414099 2017]   File "/usr/share/ipa/wsgi.py", line 51, in application
[Wed Feb 15 12:35:46.414144 2017]     return api.Backend.wsgi_dispatch(environ, start_response)
[Wed Feb 15 12:35:46.414150 2017]   File "/usr/lib/python2.7/site-packages/ipaserver/rpcserver.py", line 263, in __call__
[Wed Feb 15 12:35:46.414317 2017]     return self.route(environ, start_response)
[Wed Feb 15 12:35:46.414326 2017]   File "/usr/lib/python2.7/site-packages/ipaserver/rpcserver.py", line 275, in route
[Wed Feb 15 12:35:46.414337 2017]     return app(environ, start_response)
[Wed Feb 15 12:35:46.414340 2017]   File "/usr/lib/python2.7/site-packages/ipaserver/rpcserver.py", line 465, in __call__
[Wed Feb 15 12:35:46.414346 2017]     response = super(jsonserver, self).__call__(environ, start_response)
[Wed Feb 15 12:35:46.414349 2017]   File "/usr/lib/python2.7/site-packages/ipaserver/rpcserver.py", line 647, in __call__
[Wed Feb 15 12:35:46.414354 2017]     self.create_context(ccache=user_ccache)
[Wed Feb 15 12:35:46.414358 2017]   File "/usr/lib/python2.7/site-packages/ipalib/backend.py", line 125, in create_context
[Wed Feb 15 12:35:46.414400 2017]     time_limit=None)
[Wed Feb 15 12:35:46.414405 2017]   File "/usr/lib/python2.7/site-packages/ipalib/backend.py", line 66, in connect
[Wed Feb 15 12:35:46.414413 2017]     conn = self.create_connection(*args, **kw)
[...]

PR #112, opened 2017-03-27
Prevent GSSError/_display_status() infinite recursion

frozencemetery:
I was unable to reproduce the problem, but this should prevent the issue (which is #111).

### DIFF
--- a/gssapi/raw/misc.pyx
+++ b/gssapi/raw/misc.pyx
@@ -139,7 +139,7 @@ def _display_status(unsigned int error_code, bint is_major_code,
             whether or not to call again for further messages
 
     Raises:
-       GSSError
+       ValueError
     """
 
     cdef int status_type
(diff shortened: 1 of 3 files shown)


frozencemetery:
(Pulled some of the prep commits out of #109)

tiran:
I'm not a fan of assert() for flow control. Some flags (e.g. ```-O```) remove assert statements.

frozencemetery:
Christian Heimes <notifications@github.com> writes:

> I'm not a fan of assert() for flow control. Some flags (e.g. ```-O```)
> remove assert statements.

Understood.  Do you have a recommendation instead?  My original thought
was to raise an exception manually, and I only switched to
assert() because the assertion failure exception looked closest to the
kind of problem that was occurring.

tiran:
Any other exception will do. I like ```ValueError``` for an invalid value. You can subclass ValueError if you prefer to catch just that specific exception.

frozencemetery:
Switched to `ValueError`.  (Commit order changed too, but that's not important.)

gssapi/raw/misc.pyx
```diff
@@ -165,13 +165,15 @@ def _display_status(unsigned int error_code, bint is_major_code,
...
         gss_release_buffer(&min_stat, &msg_buff)
         return (msg_out, msg_ctx_out, call_again)
     else:
-        # NB(directxman12): this is highly unlikely to cause a recursive loop
-        raise GSSError(maj_stat, min_stat)
+        # This hides whatever error gss_display_status is complaining about,
+        # but obviates infinite recursion into stack exhaustion.  The
+        # exception raised here is handled by get_all_statuses(), which prints
+        # the code.
+        raise ValueError("gss_display_status call returned failure.")
```
DirectXMan12: please attach the error codes from `gss_display_status` into this message

frozencemetery:
Updated to include `gss_display_status` errors (for future debugging) as per request.

PR #115, opened 2017-05-31
Fix Docs Generation

DirectXMan12:
This fixes the warnings on the parsing of references to `MechType`, and fixes the bug in `requires_rfc.py` preventing us from generating documentation in newer versions of Sphinx.


frozencemetery:
Pushed to master as 7f035c1ab20b0f027631bd48c3addfa08abc68c8 and 8310103d827b3f9c40b545b35e415233a5ef7a85

PR #118, opened 2017-07-08
Update parameterized

cipherboy:
It appears that `node-parameterized` has been [deprecated](https://pypi.python.org/pypi/nose-parameterized#deprecation-warning). Updating to `parameterized` doesn't appear to break the test suite.

### DIFF
--- a/gssapi/tests/test_high_level.py
+++ b/gssapi/tests/test_high_level.py
@@ -6,7 +6,7 @@ import pickle
 
 import should_be.all  # noqa
 import six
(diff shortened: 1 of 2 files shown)


frozencemetery:
Thanks.  We will probably hold this PR back until distros start packaging parameterized instead.

DirectXMan12:
since it's a test suite dependency, it doesn't matter too much if the distros package it or not...

frozencemetery:
I'd prefer to not bring the test suite farther from being runnable on distros.

PR #121, opened 2017-07-14
Implement support for GSSAPI extension RFC 5587

cipherboy:
RFC 5587 provides extended mech inquiry calls to GSSAPI.
This adds the ability to indicate mechs by their
mech attrs, along with determining the attrs supported
by a mech. These calls are provided as a part of
the raw interface and are not exposed in the high-level
interface due to not having objects for mechs or attrs.

Thanks!


gssapi/tests/test_raw.py
```diff
@@ -644,6 +644,66 @@ def test_add_cred_with_password(self):
 
         new_creds.should_be_a(gb.Creds)
 
+    @ktu.gssapi_extension_test('rfc5587', 'RFC 5587')
+    def test_rfc5587(self):
+        mechs = gb.indicate_mechs_by_attrs(None, None, None)
+        mechs.shouldnt_be_empty()
+        mechs.should_be_a(set)
+
+        last_mech = None
```
frozencemetery: Can you just use `mechs[-1]` instead of repeatedly updating a variable for this?

gssapi/tests/test_raw.py
```diff
@@ -644,6 +644,66 @@ def test_add_cred_with_password(self):
...
+                mech_attr.should_be_a(gb.OID)
+
+                display_out = gb.display_mech_attr(mech_attr)
+                display_out.name.shouldnt_be_none
+                display_out.long_desc.shouldnt_be_none
+                display_out.short_desc.shouldnt_be_none
+
+            last_mech = mech
+
+        last_mech.shouldnt_be_none
```
frozencemetery: Isn't this check redundant with the one in the loop?

gssapi/tests/test_raw.py
```diff
@@ -644,6 +644,66 @@ def test_add_cred_with_password(self):
...
+    def test_rfc5587(self):
+        mechs = gb.indicate_mechs_by_attrs(None, None, None)
+        mechs.shouldnt_be_empty()
+        mechs.should_be_a(set)
+
+        last_mech = None
+        last_attr = None
+
+        for mech in mechs:
+            mech.shouldnt_be_none
```
frozencemetery: You need to call your functions, in this line and elsewhere

gssapi/tests/test_raw.py
```diff
@@ -644,6 +644,66 @@ def test_add_cred_with_password(self):
...
+
+                display_out = gb.display_mech_attr(mech_attr)
+                display_out.name.shouldnt_be_none
+                display_out.long_desc.shouldnt_be_none
+                display_out.short_desc.shouldnt_be_none
+
+            last_mech = mech
+
+        last_mech.shouldnt_be_none
+        last_attr.shouldnt_be_none
```
frozencemetery: Similarly, why is this checked here?

gssapi/tests/test_raw.py
cipherboy: `mechs` is a set; sets don't support indexing.

frozencemetery: So it is.  Please move the update call to the top of the loop in that case.

cipherboy:
@frozencemetery Updated!

DirectXMan12:
😢 please put a full description on the PR  body and commit message.  For example: 

```
Implement support for GSSAPI extension RFC 5587

RFC 5587 adds extended mechanism inquiry APIs to the GSSAPI.
This allows users to query mechanisms using a number of attributes associated with those
mechanisms.   This is surfaced directly in the low-level API, but is not surfaced in the
high-level API because $REASONS.
```

gssapi/raw/ext_rfc5587.pyx
```diff
@@ -0,0 +1,103 @@
...
+def inquire_attrs_for_mech(OID mech):
+    cdef OM_uint32 maj_stat, min_stat
+    cdef gss_OID m = GSS_C_NO_OID
+    cdef gss_OID_set mech_attrs = GSS_C_NO_OID_SET
+    cdef gss_OID_set known_mech_attrs = GSS_C_NO_OID_SET
+
+    if mech is not None:
+        m = &mech.raw_oid
+
+    maj_stat = gss_inquire_attrs_for_mech(&min_stat, m, &mech_attrs,
```
DirectXMan12: calls to the GSSAPI functions themselves should generally drop the GIL while being called `with nogil:`.

gssapi/raw/ext_rfc5587.pyx
```diff
@@ -0,0 +1,103 @@
...
+    cdef gss_buffer_desc long_desc
+
+    if attr is not None:
+        a = &attr.raw_oid
+
+    maj_stat = gss_display_mech_attr(&min_stat, a, &name, &short_desc,
+                                     &long_desc)
+
+    if maj_stat == GSS_S_COMPLETE:
+        out_name = name.value[:name.length].decode("UTF-8")
```
DirectXMan12: we don't decode in the low-level API -- we just return bytes.

gssapi/tests/test_raw.py
```diff
@@ -644,6 +644,64 @@ def test_add_cred_with_password(self):
...
+
+            known_mech_attrs.should_be_a(set)
+            known_mech_attrs.shouldnt_be_empty()
+
+            for mech_attr in mech_attrs:
+                mech_attr.shouldnt_be_none()
+                mech_attr.should_be_a(gb.OID)
+
+                display_out = gb.display_mech_attr(mech_attr)
+                display_out.name.shouldnt_be_none()
```
DirectXMan12: can we test a specific attr to make sure that's right, too?

gssapi/raw/ext_rfc5587.pyx
cipherboy: Ah that annotation makes sense now.  :) I notice that e.g., rfc5588 has a "nogil" on the [cdef](https://github.com/pythongssapi/python-gssapi/blob/master/gssapi/raw/ext_rfc5588.pyx#L15-L22) -- do you want that as well?

DirectXMan12: yeah, you'd need to add that

DirectXMan12:
Ok, last thing: you should have doc strings on the functions.  Otherwise, LGTM.

cipherboy:
Thanks @DirectXMan12! :)

gssapi/raw/ext_rfc5587.pyx
```diff
@@ -0,0 +1,153 @@
...
+
+    if maj_stat == GSS_S_COMPLETE:
+        return c_create_oid_set(mechs)
+    else:
+        raise GSSError(maj_stat, min_stat)
+
+
+def inquire_attrs_for_mech(OID mech):
+    """
+    inquire_attrs_for_mech(OID mech)
```
DirectXMan12: now I'm nitpicking, but the method signature in the docstring should look like normal Python (so no types in the docstring signature)

gssapi/raw/ext_rfc5587.pyx
```diff
@@ -0,0 +1,153 @@
...
+    if maj_stat == GSS_S_COMPLETE:
+        return InquireAttrsResult(c_create_oid_set(mech_attrs),
+                                  c_create_oid_set(known_mech_attrs))
+    else:
+        raise GSSError(maj_stat, min_stat)
+
+
+def display_mech_attr(OID attr):
+    """
+    display_mech_attrs(OID attr)
```
DirectXMan12: same here.

gssapi/raw/ext_rfc5587.pyx
cipherboy: Nitpicking welcome :)

cipherboy:
Are we looking for higher level changes at this time to include mechs / mech_attrs as a class of their own? It seems like they'd be a storage class with only a few helper functions, so I'm not sure if anyone's interested in having that added at this time.

cipherboy:
I've added comments to the tests and added RFC 5587 support to the README.

gssapi/tests/test_raw.py
cipherboy: Done -- see [here](https://github.com/pythongssapi/python-gssapi/pull/121/commits/66a7107c8355ca5b583480c874c30008ec9875a9#diff-5e7d4dd129f7f86ac302dc31dd036c47R719).

DirectXMan12:
LGTM.  We may have to revisit the exact message checking test in the future, but I'm fine to merge now.

PR #123, opened 2017-07-18
WIP: Channel Binding Extension

cipherboy:
This PR implements channel bindings per this draft [RFC](https://tools.ietf.org/html/draft-ietf-kitten-channel-bound-flag-01), and depends on this PR: https://github.com/krb5/krb5/pull/668. In the mean time, patches can be found in this [branch](https://github.com/cipherboy/krb5/tree/krb5-1.15.1-channel_bindings/src).

New raw calls:
  - `create_sec_context`
  - `set_context_flags`

High level changes:
  - `SecurityContext` constructor takes a `ret_flags` parameter
  - channel_bindings support -> `create_sec_context` called
  - channel_bindings support + ret_flags or req_flags set -> `set_context_flags` called


gssapi/sec_contexts.py
```diff
@@ -35,8 +38,8 @@ def __new__(cls, base=None, token=None,
 
         return super(SecurityContext, cls).__new__(cls, base)
 
-    def __init__(self, base=None, token=None,
-                 name=None, creds=None, lifetime=None, flags=None,
+    def __init__(self, base=None, token=None, name=None, creds=None,
+                 lifetime=None, flags=None, ret_flags=None,
```
DirectXMan12: this parameter name is not the best...

gssapi/sec_contexts.py
```diff
@@ -104,6 +111,17 @@ def __init__(self, base=None, token=None,
                                     "argument when creating an accepting "
                                     "security context")
 
+            if rchannel_bindings is not None:
+                empty_ctx = rchannel_bindings.create_sec_context()
+                self.copy_from(empty_ctx)
```
DirectXMan12: do this in `__new__` so you don't have to call `copy_from`.

gssapi/raw/ext_channel_bindings.pyx
```diff
@@ -0,0 +1,73 @@
...
+    init_sec_context() or accept_sec_context().
+
+    Args:
+        context (SecurityContext): Context to set flags on
+        req_flags (uint64_t): Request flags for init/accept_sec_context()
+        ret_flags_understood: Return flags that are understood by the
+            application
+
+    Returns:
+        bool: Success
```
DirectXMan12: it's fine not to return anything.  This doesn't actually return `bool: success` because it never returns `False`.

gssapi/raw/ext_channel_bindings.pyx
```diff
@@ -0,0 +1,73 @@
...
+
+    Raises:
+        GSSError
```
DirectXMan12: Is there a specific error code for in-process sec contexts?  It should be documented if there is.

gssapi/raw/ext_channel_bindings.pyx
```diff
@@ -0,0 +1,73 @@
...
+    set_context_flags(context, req_flags, ret_flags_understood)
+    Specifies request flags and understood return flags for use with
+    init_sec_context() and accept_sec_context(). Requires a context
+    previously created by create_sec_context() and not yet used with
+    init_sec_context() or accept_sec_context().
+
+    Args:
+        context (SecurityContext): Context to set flags on
+        req_flags (uint64_t): Request flags for init/accept_sec_context()
+        ret_flags_understood: Return flags that are understood by the
```
cipherboy: `ret_flags_understood` is a set of return flags that are understood by the calling application.

In other word, it is a subset of possible `ret_flag` values from `gss_init/accept_sec_context`; the only useful `ret_flag_understood` parameter to set would be the channel bound flag; setting or not setting any other flag does not change the behavior. (This last point about not changing behavior is the subject of a revision in the draft, as if you set `req_flags = GSS_C_MUTUAL_FLAG` but set `ret_flags_understood = 0`, `gss_init/accept_sec_context` will still return GSS_C_MUTUAL_FLAG on success). 

`ret_flags` was what the original [draft](https://tools.ietf.org/html/draft-ietf-kitten-channel-bound-flag-01#section-2.2.1) called it in the C bindings, hence the discrepancy in the high level API. Suggestions welcome about a more helpful name.

gssapi/raw/ext_channel_bindings.pyx
cipherboy: The only error that is possible are `GSS_S_FAILURE`, when `context.raw_ctx` is none/missing for some reason or when `context.raw_ctx` is malformed. See [here](https://github.com/cipherboy/krb5/blob/channel_bindings/src/lib/gssapi/mechglue/g_set_context_flags.c#L57).

cipherboy:
Okie dokie, this has been updated with comments. Thanks @DirectXMan12! :)

Please see the comments on some of your changes as well.

gssapi/raw/ext_channel_bindings.pyx
```diff
@@ -0,0 +1,73 @@
...
+    maj_stat = gss_create_sec_context(&min_stat, &ctx)
+
+    if maj_stat == GSS_S_COMPLETE:
+        out_ctx.raw_ctx = ctx
+        return out_ctx
+    else:
+        raise GSSError(maj_stat, min_stat)
+
+
+def set_context_flags(SecurityContext context, uint64_t req_flags,
```
DirectXMan12: don't type the flags here (since they can be `IntEnumFlagSets`), and they should either be `not None` if non-optional, or `=None` if optional (then use `or` like in https://github.com/pythongssapi/python-gssapi/blob/master/gssapi/raw/sec_contexts.pyx#L179)

gssapi/raw/sec_contexts.pyx
```diff
@@ -107,6 +107,10 @@ cdef class SecurityContext:
 
             self.raw_ctx = NULL
 
+    def copy_from(self, SecurityContext other):
```
DirectXMan12: this isn't used

gssapi/sec_contexts.py
```diff
@@ -32,11 +35,13 @@ def __new__(cls, base=None, token=None,
...
             base = rsec_contexts.import_sec_context(token)
+        elif rchannel_bindings is not None and base is None:
+            base = rchannel_bindings.create_sec_context()
 
         return super(SecurityContext, cls).__new__(cls, base)
 
-    def __init__(self, base=None, token=None,
-                 name=None, creds=None, lifetime=None, flags=None,
+    def __init__(self, base=None, token=None, name=None, creds=None,
+                 lifetime=None, flags=None, ret_flags_understood=None,
```
DirectXMan12: we *probably* should avoid changing argument order (since this isn't python 3, we can't do keyword-only args).

gssapi/sec_contexts.py
```diff
@@ -104,6 +113,14 @@ def __init__(self, base=None, token=None,
                                     "argument when creating an accepting "
                                     "security context")
 
+            if rchannel_bindings is not None:
+                i_flags = int(self._desired_flags)
+                i_ret_flags = int(self._understood_flags)
```
DirectXMan12: shouldn't need a manual cast here -- the low-level code should handle it.

gssapi/tests/test_raw.py
```diff
@@ -734,6 +734,128 @@ def test_display_mech_attr(self):
...
+                                             channel_bindings=accept_cb)
+            ctx_resp.shouldnt_be_none()
+
+            (ctx, _, _, accept_token, accept_ret_flags, _,
+             _, accept_cont_needed) = ctx_resp
+
+            if not init_cont_needed:
+                break
+
+        if ((int(init_ret_flags) & m_init_ret_flags) != e_init_ret_flags):
```
DirectXMan12: shouldn't need a cast to `int` here -- `IntEnumFlagSet` handles that case -- that's why we have it ;-).

gssapi/tests/test_raw.py
```diff
@@ -734,6 +734,128 @@ def test_display_mech_attr(self):
...
+            [bad_bindings, f_both, 0, 0, None, cb, f_cb, 0, f_cb, 0],
+            [0, f_both, 0, 0, cb, cb, f_cb, 0, f_cb, f_cb],
+            [0, f_both, 0, 0, cb, None, f_cb, 0, f_cb, 0],
+            [0, f_both, f_cb, f_cb, cb, None, f_cb, 0, f_cb, 0],
+            [0, f_both, f_cb, f_cb, None, None, f_cb, 0, f_cb, 0],
+            [0, f_both, f_cb, f_cb, None, cb, f_cb, 0, f_cb, 0],
+            [0, f_both, f_cb, f_cb, cb, cb, f_cb, 0, f_cb, f_cb],
+        ]
+
+        for test in tests:
```
DirectXMan12: use parameterized?

PR #124, opened 2017-07-24
Implement support for GSSAPI extension RFC 5801

cipherboy:
[RFC 5801](https://tools.ietf.org/html/rfc5801) provides SASL-specific mechanism inquiry
calls for GSSAPI. The inquire_mech_for_saslname
call allows resolving mechs by their respective
SASL names, and inquire_saslname_for_mech allows
the reverse operation, also providing a mechanism
name and description. These calls are implemented
as part of the raw interface.


cipherboy:
@DirectXMan12 / @frozencemetery Would you like specific tests on this one as well? They'd again have to be implementation-specific, as the RFC doesn't specify OID<->SASL Name mappings and/or mech names/descriptions.

frozencemetery:
I don't particularly want implementation-specific tests, but will let @DirectXMan12 decide that.

gssapi/raw/ext_rfc5801.pyx
```diff
@@ -0,0 +1,88 @@
...
+        gss_buffer_t mech_name,
+        gss_buffer_t mech_description) nogil
+
+    OM_uint32 gss_inquire_mech_for_saslname(
+        OM_uint32 *min_stat,
+        const gss_buffer_t sasl_mech_name,
+        gss_OID *mech_type) nogil
+
+
+def inquire_saslname_for_mech(OID mech):
```
DirectXMan12: RFC 5801 doesn't mark `mech` as optional, so this should be marked `not None` -- it shouldn't be an option to pass no mech.

cipherboy:
Updated. Thanks @DirectXMan12! :)

gssapi/raw/ext_rfc5801.pyx
```diff
@@ -0,0 +1,88 @@
...
+    inquire_saslname_for_mech(mech)
+    Gets information about a specified mech, including the SASL name,
+    the mech name, and the mech description.
+
+    Args:
+        mech (OID): Mechanism to inquire about
+
+    Returns:
+        InquireSASLNameResult: the results of inquiry; a mech's SASL name,
+            name, and description.
```
DirectXMan12: missing a `Raises` clause.

gssapi/raw/ext_rfc5801.pyx
```diff
@@ -0,0 +1,88 @@
...
+def inquire_mech_for_saslname(bytes sasl_name):
+    """
+    inquire_mech_for_saslname(sasl_name)
+    Gets the OID for the mech specified by SASL name.
+
+    Args:
+        sasl_name (bytes): SASL name of the mechanism
+
+    Returns:
+        OID: the mechanism with corresponding SASL name.
```
DirectXMan12: missing a `Raises` clause.

gssapi/raw/ext_rfc5801.pyx
```diff
@@ -0,0 +1,88 @@
...
+        InquireSASLNameResult: the results of inquiry; a mech's SASL name,
+            name, and description.
+    """
+    cdef OM_uint32 maj_stat, min_stat
+    cdef gss_buffer_desc sasl_mech_name
+    cdef gss_buffer_desc mech_name
+    cdef gss_buffer_desc mech_desc
+    cdef gss_OID m = GSS_C_NO_OID
+
+    if mech is not None:
```
DirectXMan12: no need for this, because you have the `not None` clause above.

gssapi/raw/ext_rfc5801.pyx
```diff
@@ -0,0 +1,88 @@
...
+        gss_release_buffer(&min_stat, &sasl_mech_name)
+        gss_release_buffer(&min_stat, &mech_name)
+        gss_release_buffer(&min_stat, &mech_desc)
+
+        return InquireSASLNameResult(out_smn, out_mn, out_md)
+    else:
+        raise GSSError(maj_stat, min_stat)
+
+
+def inquire_mech_for_saslname(bytes sasl_name):
```
DirectXMan12: this should also be `not None`, presumably.

DirectXMan12:
Will merge once we get Travis working

PR #125, opened 2017-07-25
Fix OID inequality comparison

cipherboy:
Due to an extraneous 'or' branch, OID inequality
comparison would return the results of an equality
comparison.

### DIFF
--- a/gssapi/tests/test_raw.py
+++ b/gssapi/tests/test_raw.py
@@ -1353,3 +1353,13 @@ class TestOIDTransforms(unittest.TestCase):
             int_seq = oid['string'].split('.')
             o = gb.OID.from_int_seq(int_seq)
             o.__bytes__().should_be(oid['bytes'])
+
+    def test_comparisons(self):
+        krb5 = gb.OID.from_int_seq(TEST_OIDS['KRB5']['string'])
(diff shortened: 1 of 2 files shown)


gssapi/tests/test_raw.py
```diff
@@ -1353,3 +1353,20 @@ def test_encode_from_int_seq(self):
             int_seq = oid['string'].split('.')
             o = gb.OID.from_int_seq(int_seq)
             o.__bytes__().should_be(oid['bytes'])
+
+    def test_comparisons(self):
+        krb5 = gb.OID.from_int_seq(TEST_OIDS['KRB5']['string'])
+        krb5_other = gb.OID.from_int_seq(TEST_OIDS['KRB5']['string'])
+        spnego = gb.OID.from_int_seq(TEST_OIDS['SPNEGO']['string'])
+
+        a = krb5 == krb5_other
```
DirectXMan12: make this `(krb5 == krb5_other).should_be(True)`  so that we get nice error messages with should_be

DirectXMan12:
Good catch.

/me is embarrased

cipherboy:
Thank you! Updated :)

PR #126, opened 2017-07-25
Expose mechanisms in the high-level API

cipherboy:
This creates a new class, Mechanism, for inquiring
information about a mechanism. This includes support
for RFCs 5587 and 5801. As Mechanism derives from OID,
it is compatible with all places that accept a mech
by OID.

Points of discussion (mechs):
  - ~Fallback in `mech_name()`: what mechs should we support? We could hard-code the three MIT/KRB5 OIDs (krb5/iakerb/spnego) with respective names for the fallback when RFC5801 is missing, but I'd like thoughts on that.~
  - ~Am I missing functions that significantly depend on a OID?~
  - Do we want other high-level calls to default to exposing Mechanisms vs. raw OIDs? e.g., `init_sec_context`?
  - ~Revising __init__() -- I wasn't able to get a call to `super().__init__` to work like was done with creds.   Thoughts would be appreciated.~

Regarding mech_attrs, I'm not yet sure how I want to handle that. `display_mech_attr` is about the only useful function I can see that depends on mech_attrs, and querying for a list of mech_attrs isn't trivial--It'd likely involve a call to `indicate_mechs`+`inquire_attrs_for_mechs`. I'm not really seeing a use for having mech_attrs as a class. Thoughts?

### DIFF
--- a/gssapi/mechs.py
+++ b/gssapi/mechs.py
@@ -1,0 +1,199 @@
+import six
+
+from gssapi.raw import oids as roids
+from gssapi._utils import import_gssapi_extension
+from gssapi.raw import misc as rmisc
+from gssapi import _utils
+
+rfc5587 = import_gssapi_extension('rfc5587')
+rfc5801 = import_gssapi_extension('rfc5801')
+
+
+class Mechanism(roids.OID):
+    """
+    A GSSAPI Mechanism
+
+    This class represents a mechanism and centralizes functions dealing with
+    mechanisms and can be used with any calls.
+
+    It inherits from the low-level GSSAPI :class:`~gssapi.raw.oids.OID` class,
+    and thus can be used with both low-level and high-level API calls.
+    """
+    def __new__(cls, cpy=None, elements=None):
+        return super(Mechanism, cls).__new__(cls, cpy, elements)
+
+    @property
+    def name_types(self):
+        """
+        Get the set of name types supported by this mechanism.
+        """
+        return rmisc.inquire_names_for_mech(self)
+
+    @property
+    def _saslname(self):
+        if rfc5801 is None:
+            raise NotImplementedError("Your GSSAPI implementation does not "
+                                      "have support for RFC 5801")
+        return rfc5801.inquire_saslname_for_mech(self)
+
+    @property
+    def _attrs(self):
+        if rfc5587 is None:
+            raise NotImplementedError("Your GSSAPI implementation does not "
+                                      "have support for RFC 5587")
+
+        return rfc5587.inquire_attrs_for_mech(self)
+
+    def __str__(self):
+        if issubclass(str, six.text_type):
+            # Python 3 -- we should return unicode
+            return self._bytes_desc().decode(_utils._get_encoding())
+        else:
+            return self._bytes_desc()
+
+    def __unicode__(self):
+        return self._bytes_desc().decode(_utils._get_encoding())
+
+    def _bytes_desc(self):
+        base = self.dotted_form
+        if rfc5801 is not None:
+            base = self._saslname.mech_name
+
+        if isinstance(base, six.text_type):
+            base = base.encode(_utils._get_encoding())
+
+        return base
+
+    def __repr__(self):
+        """
+        Get a name representing the mechanism; always safe to call
+        """
+        base = "<Mechanism (%s)>" % self.dotted_form
+        if rfc5801 is not None:
+            base = "<Mechanism %s (%s)>" % (
+                self._saslname.mech_name.decode('UTF-8'),
+                self.dotted_form
+            )
+
+        return base
+
+    @property
+    def sasl_name(self):
+        """
+        Get the SASL name for the mechanism
+
+        :requires-ext:`rfc5801`
+        """
+        return self._saslname.sasl_mech_name.decode('UTF-8')
+
+    @property
+    def description(self):
+        """
+        Get the description of the mechanism
+
+        :requires-ext:`rfc5801`
+        """
+        return self._saslname.mech_description.decode('UTF-8')
+
+    @property
+    def known_attrs(self):
+        """
+        Get the known attributes of the mechanism; returns a set of OIDs
+        ([OID])
+
+        :requires-ext:`rfc5587`
+        """
+        return self._attrs.known_mech_attrs
+
+    @property
+    def attrs(self):
+        """
+        Get the attributes of the mechanism; returns a set of OIDs ([OID])
+
+        :requires-ext:`rfc5587`
+        """
+        return self._attrs.mech_attrs
+
+    @classmethod
+    def all_mechs(cls):
+        """
+        Get a generator of all mechanisms supported by GSSAPI
+        """
+        return (cls(mech) for mech in rmisc.indicate_mechs())
+
+    @classmethod
+    def from_name(cls, name=None):
+        """
+        Get a generator of mechanisms that may be able to process the name
+
+        Args:
+            name (Name): a name to inquire about
+
+        Returns:
+            [Mechanism]: a set of mechanisms which support this name
+
+        Raises:
+            GSSError
+        """
+        return (cls(mech) for mech in rmisc.inquire_mechs_for_name(name))
+
+    @classmethod
+    def from_sasl_name(cls, name=None):
+        """
+        Create a Mechanism from its SASL name
+
+        Args:
+            name (str): SASL name of the desired mechanism
+
+        Returns:
+            Mechanism: the desired mechanism
+
+        Raises:
+            GSSError
+
+        :requires-ext:`rfc5801`
+        """
+        if rfc5801 is None:
+            raise NotImplementedError("Your GSSAPI implementation does not "
+                                      "have support for RFC 5801")
+        if isinstance(name, six.text_type):
+            name = name.encode(_utils._get_encoding())
+
+        m = rfc5801.inquire_mech_for_saslname(name)
+
+        return cls(m)
+
+    @classmethod
+    def from_attrs(cls, m_desired=None, m_except=None, m_critical=None):
+        """
+        Get a generator of mechanisms supporting the specified attributes. See
+        RFC 5587's :func:`indicate_mechs_by_attrs` for more information.
+
+        Args:
+            m_desired ([OID]): Desired attributes
+            m_except ([OID]): Except attributes
+            m_critical ([OID]): Critical attributes
+
+        Returns:
+            [Mechanism]: A set of mechanisms having the desired features.
+
+        Raises:
+            GSSError
+
+        :requires-ext:`rfc5587`
+        """
+        if isinstance(m_desired, roids.OID):
+            m_desired = set([m_desired])
+        if isinstance(m_except, roids.OID):
+            m_except = set([m_except])
+        if isinstance(m_critical, roids.OID):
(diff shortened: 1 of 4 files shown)


gssapi/mechs.py
```diff
@@ -0,0 +1,180 @@
...
+        """
+        if rfc5801 is None:
+            return str(roids.OID(self))
+        self._query_saslname()
+        return self._sasl_name.mech_name.decode('UTF-8')
+
+    @property
+    def sasl_name(self):
+        """
+        Get the SASL name for the mechanism; depends on RFC 5801
```
cipherboy: (done)

cipherboy:
Thanks for the detailed review and the help tracking those issues down yesterday. :)

I added `inquire_mechs_for_name()` support as a missing mechanism-related function.

TODO:
  - Python 2/3 `str` handling.
  - Fallback mech names for pretty printing
  - `mech_attrs`

frozencemetery:
Hardcoding anything with a standard is fine in principle, but kind of defeats the point of calling into the GSSAPI there at all... I'm not decided on that.

cipherboy:
On second thought, I don't know if there is value to this. Both MIT and Heimdal support the RFC 5801 extensions, so mech_name will be available under both.

gssapi/mechs.py
```diff
@@ -0,0 +1,222 @@
...
+        return self._query_saslname.mech_name.decode('UTF-8')
+
+    @property
+    def description(self):
+        """
+        Get the description of the mechanism; depends on RFC 5801
+
+        :requires-ext:`rfc5801`
+        """
+        self._query_saslname()
```
DirectXMan12: redundant call

gssapi/mechs.py
```diff
@@ -0,0 +1,222 @@
...
+        return self._query_attrs.mech_attrs
+
+    @classmethod
+    def all_mechs(cls):
+        """
+        all_mechs()
+        Get a list of all mechanisms supported by GSSAPI
+        """
+        mechs = rmisc.indicate_mechs()
+        return [cls(mech) for mech in mechs]
```
DirectXMan12: collapse to `(cls(mech) for mech in metchs)`

gssapi/mechs.py
```diff
@@ -0,0 +1,222 @@
...
+        Raises:
+            GSSError
+
+        :requires-ext:`rfc5801`
+        """
+        if rfc5801 is None:
+            raise NotImplementedError("Your GSSAPI implementation does not "
+                                      "have support for RFC 5801")
+        n = name
+        if type(n) is not bytes:
```
DirectXMan12: use `isinstance`

gssapi/mechs.py
```diff
@@ -0,0 +1,205 @@
...
+
+    def __bytes__(self):
+        base = self.dotted_form
+        if rfc5801 is not None:
+            base = self._saslname.mech_name
+
+        if issubclass(str, six.text_type):
+            base = bytes(base, _utils._get_encoding())
+        else:
+            base = bytes(base)
```
cipherboy: Since `dotted_form` returns a string, this needs to be cast to `bytes` for python2 compatibility. However, python3 forces two arguments to `bytes`, one of which is the encoding. Suggestions on how to make this better?

gssapi/raw/oids.pyx
```diff
@@ -120,11 +132,15 @@ cdef class OID:
         if self._free_on_dealloc:
             free(self.raw_oid.elements)
 
-    def __bytes__(self):
+    @property
+    def byte_form(self):
```
DirectXMan12: err... why is this necessary?  Why not just leave it as `__bytes__`?

gssapi/mechs.py
```diff
@@ -0,0 +1,205 @@
...
+        if issubclass(str, six.text_type):
+            # Python 3 -- we should return unicode
+            return self.__bytes__().decode(_utils._get_encoding())
+        else:
+            return self.__bytes__()
+
+    def __unicode__(self):
+        return self.__bytes__().decode(_utils._get_encoding())
+
+    def __bytes__(self):
```
DirectXMan12: hmm... I'd just leave `__bytes__` as-is.  Let in inherit from the low-level API for compatibility with OIDs -- some things might expect OID bytes when calling `__bytes__` -- we shouldn't change that part of the API just because you're working Mechanisms.

DirectXMan12: then put the "string form as bytes" in a separate private method (like `_byte_desc`).

cipherboy:
Okie dokie, this is updated as well. In retrospect, not changing the `__bytes__` interface was probably a good idea. :)

gssapi/mechs.py
DirectXMan12: it should be fine as

```python
if isinstance(base, six.text_type):
    base = base.encode(_utils._get_encoding())
```

gssapi/mechs.py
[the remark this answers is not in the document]
DirectXMan12: while you're doing updates, remove the "; depends on RFC 5801" -- just the `requires-ext` tag below is fine.

gssapi/mechs.py
```diff
@@ -0,0 +1,205 @@
...
+            GSSError
+
+        :requires-ext:`rfc5801`
+        """
+        if rfc5801 is None:
+            raise NotImplementedError("Your GSSAPI implementation does not "
+                                      "have support for RFC 5801")
+        n = name
+        if not isinstance(n, bytes):
+            n = str(n).encode()
```
DirectXMan12: The pattern we generally use here is

```python
if isinstance(name, six.text_type):
    name = name.encode(_utils._get_encoding())
```

See the `gssapi/names.py` (at the bottom) for example

gssapi/tests/test_high_level.py
```diff
@@ -369,6 +370,57 @@ def test_add_with_impersonate(self):
...
+            s.shouldnt_be_empty()
+
+    @ktu.gssapi_extension_test('rfc5801', 'RFC 5801: SASL Names')
+    def test_sasl_properties(self):
+        mechs = gssmechs.Mechanism.all_mechs()
+        for mech in mechs:
+            s = str(mech)
+            s.shouldnt_be_empty()
+            s.should_be_a(str)
+            s[0].shouldnt_be('<')
```
DirectXMan12: this is a fairly fragile line here.  I'd just remove it, unless the SASL names are *guaranteed* to not begin with "<"...

gssapi/tests/test_high_level.py
```diff
@@ -369,6 +370,57 @@ def test_add_with_impersonate(self):
...
+        for mech in mechs:
+            attrs = mech.attrs
+            known_attrs = mech.known_attrs
+
+            for attr in attrs:
+                i = list(gssmechs.Mechanism.from_attrs(m_desired=[attr]))
+                e = list(gssmechs.Mechanism.from_attrs(m_except=[attr]))
+
+                count = len(i) + len(e)
+                count.should_be(c)
```
DirectXMan12: `(len(from_desired) + len(from_except)).should_be(c)` so we get a good error message like `len(from_desired) + len(from_except) should be 37`.

gssapi/raw/oids.pyx
```diff
@@ -61,6 +61,9 @@ cdef class OID:
         self._free_on_dealloc = True
         return 0
 
+    def _copy_oid(self, OID other):
```
DirectXMan12: this doesn't seem to be needed any more

gssapi/mechs.py
```diff
@@ -0,0 +1,202 @@
...
+    mechanisms and can be used with any calls.
+
+    It inherits from the low-level GSSAPI :class:`~gssapi.raw.oids.OID` class,
+    and thus can be used with both low-level and high-level API calls.
+    """
+    def __new__(cls, cpy=None, elements=None):
+        return super(Mechanism, cls).__new__(cls, cpy, elements)
+
+    @property
+    def names(self):
```
DirectXMan12: this should probably be `name_types`, to be clearer -- we use the term `name_type` elsewhere in the high-level API

gssapi/mechs.py
```diff
@@ -0,0 +1,202 @@
...
+        Get the SASL name for the mechanism
+
+        :requires-ext:`rfc5801`
+        """
+        return self._saslname.sasl_mech_name.decode('UTF-8')
+
+    @property
+    def description(self):
+        """
+        Get the description of the mechanism; depends on RFC 5801
```
DirectXMan12: you can remove `depends on RFC 5801`, since you have the `requires-ext` tag.

gssapi/mechs.py
```diff
@@ -0,0 +1,202 @@
...
+
+        :requires-ext:`rfc5801`
+        """
+        return self._saslname.mech_description.decode('UTF-8')
+
+    @property
+    def known_attrs(self):
+        """
+        Get the known attributes of the mechanism; depends on RFC 5587
```
DirectXMan12: You should document what form an "attribute" takes.

gssapi/mechs.py
DirectXMan12: ditto for the cases below

gssapi/mechs.py
```diff
@@ -0,0 +1,202 @@
...
+        Get the attributes of the mechanism; depends on RFC 5587
+
+        :requires-ext:`rfc5587`
+        """
+        return self._attrs.mech_attrs
+
+    @classmethod
+    def all_mechs(cls):
+        """
+        all_mechs()
```
DirectXMan12: no need for the explicit signature here.

DirectXMan12: ditto below

gssapi/mechs.py
```diff
@@ -0,0 +1,202 @@
...
+        m = rfc5801.inquire_mech_for_saslname(name)
+
+        return cls(m)
+
+    @classmethod
+    def from_attrs(cls, m_desired=None, m_except=None, m_critical=None):
+        """
+        from_attrs
+        Get a generator of mechanisms supporting the specified attributes. See
+        RFC 5587's indicate_mechs_by_attrs for more information.
```
DirectXMan12: link to `indicate_mechs_by_attrs` using the correct documentation tag (IIRC, it's something like `:func:`).

gssapi/mechs.py
cipherboy: Sorry, should've caught those sooner if I were thinking when I updated the PR.

gssapi/mechs.py
```diff
@@ -0,0 +1,199 @@
...
+        ([OID])
+
+        :requires-ext:`rfc5587`
+        """
+        return self._attrs.known_mech_attrs
+
+    @property
+    def attrs(self):
+        """
+        Get the attributes of the mechanism; returns a set of OIDs ([OID])
```
cipherboy: Leaving this open as I'm not sure if this is the desired format.

cipherboy: Per IRC discussion with @DirectXMan12 this appears to be in the correct format.

cipherboy:
@DirectXMan12 -- any new feedback would be appreciated. :)

DirectXMan12:
/me attempts to page this back into RAM

DirectXMan12:
Looks like my last comments were just around docs, so assuming past me's review was fairly thorough (I recall it being so), I'm going to merge this.

PR #127, opened 2017-09-05
Fix Fedora rawhide builders

frozencemetery:
This seems to happen roughly every time it branches; see build log for #124.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -25,19 +25,22 @@ if [ -f /etc/debian_version ]; then
     source ./.venv/bin/activate
 elif [ -f /etc/redhat-release ]; then
     # yum has no update-only verb
+    function yuminst() {
+        yum -y --nogpgcheck install $@
+    }
 
-    yum -y install krb5-{devel,libs,server,workstation} which gcc findutils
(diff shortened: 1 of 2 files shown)


DirectXMan12:
seems reasonable-ish.  Would prefer to have GPG checks, but if they break when fedora branches, I'm not sure there's much we can do about it...

frozencemetery:
Agreed... it's probably not something we want in our release builds if we can avoid it, but on the other hand, release builds aren't working right now and haven't been for a while.

frozencemetery:
@DirectXMan12 Please review added commit.  Thanks!

DirectXMan12:
can you put the RFC number after `is permitted by RFC` (I looked at 2743/2744, but couldn't find exactly what you were referring to), and also put a rough version near `in newever versions` for posterity?  Otherwise, looks good.

DirectXMan12:
(also possibly add a more elaborate comment than `# clockskew`)

frozencemetery:
Added version to commit and clarified language.  Elaborated on comment inline.

PR #133, opened 2017-11-29
Fix Travis & Travis Build Stages

DirectXMan12:
This fixes our Travis auto-deployment mechnanisms, and switches
over to making use of Travis build stages along the way.

There are now three stages:

1. verify: this *just* runs the verification suite, which currently
   consists of the flake8 tests.  The verification is run under
   both Python 2 and Python 3, using the fedora:rawhide image

2. test: this runs all of non-verification tests, under the full
   test matrix we were using before

3. deploy: this runs our deployements, as before

The deploy phase now precompiles the sdist, and asks for that to be
uploaded instead of regenerating the code before upload.  This is a bit
of a hack.

Finally, the actual Travis scripts themselves were refactored into
a series of smaller files and utility functions, which makes the code
a bit more readable in some cases, and allows individual functions
to be called separately (useful in testing stuff out, and in creating
new build stages).

This PR also fixes a verification failure, and switches the `tox` verification
over to using the lib-verify functions created for Travis.


DirectXMan12:
@frozencemetery 🎉 

this should work, I believe.  There's one last thing to verify, once you review this.

Also, I have some docker-related optimizations in the wings that go on top of this code, but I wasn't able to quite finish them tonight (the gist is that we'd build Docker images on a nightly basis using Travis cron jobs for each item in our build matrix, so that we only have to install the project dependencies in the docker container, which could speed things up a bit)

tox.ini
```diff
@@ -9,10 +9,9 @@ envlist = py27,py33,py34,py35
 [testenv]
 # NB(sross): disabling E225,E226,E227,E901 make pep8 think Cython is ok
 # NB(sross): disable E402 because of the GSSAPI="blah" lines needed by Cython
```
frozencemetery: These comments should be moved to where they live now.  (They're duplicated there, so I think they should go away entirely.)

.travis/lib-setup.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
```
frozencemetery: shebang with `-ex` so we see failures and trace from this.

.travis/lib-setup.sh
```diff
@@ -0,0 +1,87 @@
...
+    if [ -f /etc/fedora-release ]; then
+        setup::fedora::install
+    else
+        setup::centos::install
+    fi
+}
+
+setup::install() {
+    if [ -f /etc/debian_version ]; then
+        setup::debian::install
```
frozencemetery: I'm not sold on these being their own functions, but if you are then it's fine.

.travis/lib-util.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
frozencemetery: Shebang with -ex.

.travis/lib-verify.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
frozencemetery: Shebang with -ex.

.travis/lib-verify.sh
```diff
@@ -0,0 +1,25 @@
...
+
+verify::flake8() {
```
frozencemetery: Would it be better to move all the lib-*.sh functions into one file? I'm not sold on having this many files, especially when a few of them only have a single function in them.

tox.ini
DirectXMan12: ack, I just forgot to remove them.

.travis/lib-verify.sh
DirectXMan12: I just like splitting things up logically (I almost prefer smaller bash files anyway -- they're easier to manage)

PR #138, opened 2017-12-20
Fix typo.

erinn:
Small typo fix.


frozencemetery:
Thanks for noticing!

(I edited your commit slightly rather than asking you to address the nits, but basically - more descriptive message is helpful, and reflow text.)

Issue #51, opened 2015-02-16
Implement Global Grid Forum Extensions


DirectXMan12:
- gss_inquire_sec_context_by_oid
- gss_inquire_cred_by_oid
- gss_set_sec_context_option
- gss_set_cred_option (probably not GGF)
- gss_{import,export}_cred (GGF variants, nobody implements them, so we should just skip them)

jborean93:
@DirectXMan12 the docs seem to be out of date but I cannot find any reference from the GFF doc https://www.ogf.org/documents/GFD.24.pdf and `gss_set_cred_option`. Then again the documentation on the GFF extensions seem to be very rare so I could be missing something.

frozencemetery:
@jborean93 Hmm, it's not mentioned in the [IETF draft](https://tools.ietf.org/html/draft-engert-ggf-gss-extensions-00) either.  We pulled this list from [krb5's gssapi_ext.h](https://github.com/krb5/krb5/blob/master/src/lib/gssapi/generic/gssapi_ext.h#L131-L182) I think, which does have it there.  Doing some excavation it doesn't look like it is actually a GGF extension:

- It says GGF extension in gssapi_ext.h, but it's not in any GGF docs I can find
- at 19b34d5112cc13214f6a47962be89f199966a449 it had a note of /* XXX do these really belong in this header? */ right above it                         
- looks like it was gssspi before that?
- heimdal added it in 2006, with a 2004 copyright on PADL and no other notes
- krb5 adds it in a huge (svn) merge commit in 2009

@DirectXMan12 do you have thoughts on this?

frozencemetery:
Upstream believes it's a nonstandard extension mostly from Heimdal (possibly with MIT input).  In light of that, it doesn't belong with the other GGF extensions, and the only documentation we're going to get is likely to be the implementations.

frozencemetery:
Further, it appears that gss_import/export_cred are from the GGF extensions, so we probably want to pull that with these.

jborean93:
@frozencemetery I saw references to that but it seems like there is already in implementation for a different gss_import/export_cred here https://github.com/pythongssapi/python-gssapi/blob/master/gssapi/raw/ext_cred_imp_exp.pyx which is

```
OM_uint32 gss_export_cred(
        OM_uint32 *min_stat,
        gss_cred_id_t cred_handle,
        gss_buffer_t token)
```

This is different from the GGF definition

```
OM_uint32 gss_export_cred(
[...]

PR #128, opened 2017-09-25
Implement support for GSSAPI extension gss_inquire_sec_context_by_oid

vm86:
Hello, I added a raw function : gss_inquire_sec_context_by_oid .

From this issues:
https://github.com/pythongssapi/python-gssapi/issues/51

was tested on the SMB protocol.

PR #141, opened 2018-01-11
Added support for GGF inquire cred and sec context

jborean93:
Implements gss_inquire_cred_by_oid and gss_inquire_sec_context_by_oid. This partially completes https://github.com/pythongssapi/python-gssapi/issues/51.

I was trying to also implement the other 2 methods and wasn't able to get a working implementation so have decided to ignore it for now.


jborean93:
This is also a superset of https://github.com/pythongssapi/python-gssapi/pull/128 but also includes some documentation and tests as well as the `inquire_cred_by_oid` method

jborean93:
Thanks @frozencemetery will do

jborean93:
@frozencemetery let me know if the latest commit message is what you are looking for.

vm86:
Thanks! I unfortunately did not have time.

gssapi/raw/ext_ggf.pyx
```diff
@@ -0,0 +1,114 @@
...
+
+    This method inspects a :class:`Creds` object for information
+    specific to a particular mechanism.
+
+    Args:
+        cred_handle (Creds): the security context to query
+        mech (OID): the desired mechanism
+
+    Returns:
+        list: A list of zero or more pieces of data corresponding to the
```
DirectXMan12: nit: a list of zero or more bytes objects?  Can you be more specific about what a piece of data is?

gssapi/raw/ext_ggf.pyx
```diff
@@ -0,0 +1,114 @@
...
+def inquire_cred_by_oid(Creds cred_handle not None, OID mech not None):
+    """
+    inquire_cred_by_oid(cred_handle, mech)
+
+    This method inspects a :class:`Creds` object for information
+    specific to a particular mechanism.
+
+    Args:
+        cred_handle (Creds): the security context to query
+        mech (OID): the desired mechanism
```
DirectXMan12: nit: is this really a mechanism OID, or is it just some other type of OID?

DirectXMan12:
Also, could you put that link and info from the commit message into a module-level doc string?

gssapi/raw/ext_ggf.pyx
jborean93: Not sure what the terms actually mean but this is probably more `desired_object (OID): the OID of the desired object` or something like that?

gssapi/raw/ext_ggf.pyx
jborean93: I'm honestly not really sure, I've only used this to retrieve `GSS_C_INQ_SSPI_SESSION_KEY` and the end result in a list with a single entry of a byte string. I'm happy to say the "piece of data" is a byte string but not sure what else it could be.

gssapi/raw/ext_ggf.pyx
DirectXMan12: yeah, that's fine (alternatively, something like `desired_aspect (OID): the desired aspect of the Credentials to inquire about` (also, looks like you have have a copy-paste error above with creds vs contexts)

gssapi/raw/ext_ggf.pyx
simo5: @DirectXMan12 no you can't be more precise, because the data is a bunch of bytes, and how you interpret them depends on the desired oid used.

DirectXMan12: I mean, if the actual return type is just multiple chunks of bytes, then it's fine to say

> a list of zero or more pieces of data (as `bytes` objects)

or something of the sort

jborean93:
@DirectXMan12 so just add the commit message to a docstring at the top of the pyx file?

DirectXMan12:
Something like that should work, yes.  When people look at the GGF extension module and they're wondering what it's all about, they'll see that documentation, so it should serve that purpose, more or less.

jborean93:
No worries, will update the PR now. Thanks for the review.

gssapi/raw/ext_ggf.pyx
```diff
@@ -1,3 +1,16 @@
+"""
+GGF Extensions
+
+GGF provides extended credential and security context inquiry that allows
+application to retrieve more information about the client's credentials and
+security context. One common use case is to use gss_inquire_sec_context_by_oid
```
DirectXMan12: should just be `inquire_sec_context_by_oid`, and you should be able to make it a link to the actual function documentation below.

gssapi/raw/ext_ggf.pyx
```diff
@@ -1,3 +1,16 @@
...
+to retrieve the "session" key that is required by the SMB protocol for signing
+and encrypting a message. These calls are provided as a part of the raw
+interface and are not exposed in the high-level interface.
```
DirectXMan12: no need for the sentence starting with "These calls".  It just means if we ever expose it at the high level, we'll have to remember to remove that sentence.

gssapi/raw/ext_ggf.pyx
jborean93: Do I link it by enclosing it with `? Sorry just can't find an example.

jborean93:
Thanks @DirectXMan12 hopefully the latest changes are what you are looking for.

jborean93:
sigh Travis-CI looks like it is still having issues, trying to run the tests again.

jborean93:
Thanks @DirectXMan12 @frozencemetery for the reviews and help.

PR #142, opened 2018-01-11
allow setup to override the krb5-config path

jborean93:
This change allows a user to override the default path used to to determine the krb5 install prefix if they don't want to default system one to be used. I had to override this so that I could get python-gssapi to compile with the MIT krb5 install I had on MacOS compared to the default Heimdal one which doesn't support some of the extensions I had.

Side note if you know of a better way of doing this please let me know.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,7 +65,8 @@ link_args = link_args.split()
 compile_args = compile_args.split()
 
 # add in the extra workarounds for different include structures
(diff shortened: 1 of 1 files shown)


jborean93:
On testing again I mustn't have reloaded my bash profile when initially testing it, subprocess is picking up the custom `krb5-config` if I overrode it in the path. Going to close the PR as it shouldn't be needed.

Issue #119, opened 2017-07-11
Travis OSX support


frozencemetery:
It would be good, since we at some point did successfully work on OSX, to have CI running against it.  I believe Travis has support for this.

PR #146, opened 2018-03-01
Added MacOS to Travis-CI build tests

jborean93:
Added basic build verification when running on a MacOS host. Doesn't actually run the tests themselves but will verify python-gssapi will compile and install.

Partially covers https://github.com/pythongssapi/python-gssapi/issues/119

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,15 +6,7 @@ language: python
 services:
   - docker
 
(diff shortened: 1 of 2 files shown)


DirectXMan12:
Neat, thanks!.  Now we just need to get the tests working...

PR #150, opened 2018-03-21
Handle GSS_C_NO_OID_SET when creating sets

### LINKED ISSUE #148: gssapi.raw.inquire_attrs_for_mech may segfault
When using this function gss_inquire_attrs_for_mech() is called.
This function can legally return GSS_S_COMPLETE with GSS_NO_OID_SET for both return arguments.
However we do not catch this case and try to build sets for both causing a seggault as soon as we try to check for mech_set.count inside of c_create_oid_set()

You can test this with MIT (any version) and GSS-NTLMSSP and this snippet:
```
import gssapi

ntlm_mech = gssapi.OID.from_int_seq("1.3.6.1.4.1.311.2.2.10")
gssapi.raw.inquire_attrs_for_mech(ntlm_mech)
```

DirectXMan12:
Some methods can return GSS_C_NO_OID_SET on success, so we should handle
that in our set converter by returning the empty set.

Fixes #148

### DIFF
--- a/gssapi/raw/cython_converters.pyx
+++ b/gssapi/raw/cython_converters.pyx
@@ -29,6 +29,13 @@ cdef gss_OID_set c_get_mech_oid_set(object mechs):
 cdef object c_create_oid_set(gss_OID_set mech_set, bint free=True):
     """Convert a GSS OID set struct to a set of OIDs"""
 
+    if mech_set == GSS_C_NO_OID_SET:
+        # return the empty set if the we get passed the C equivalent
+        # (it could be argued that the C equivalent is closer to None,
+        # but returning None would make the API harder to work with,
+        # without much value)
+        return set()
+
     py_set = set()
     cdef i
     for i in range(mech_set.count):
--- a/.travis/lib-setup.sh
+++ b/.travis/lib-setup.sh
@@ -20,7 +20,8 @@ setup::debian::install() {
     if [ x"$KRB5_VER" = "xheimdal" ]; then
         apt-get -y install heimdal-dev
     else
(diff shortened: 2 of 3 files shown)


gssapi/raw/cython_converters.pyx
```diff
@@ -29,6 +29,13 @@ cdef gss_OID_set c_get_mech_oid_set(object mechs):
 cdef object c_create_oid_set(gss_OID_set mech_set, bint free=True):
     """Convert a GSS OID set struct to a set of OIDs"""
 
+    if mech_set == GSS_C_NO_OID_SET:
+        # return the empty set if the we get passed the C equivalent
+        # (it could be argued that the C equivalent is closer to None,
+        # but returning None would make the API harder to work with,
+        # without much value)
+        return set()
+
```
simo5: If we ever need the distinction we may need to make an incompatible API change and start returning None, is that ok ?

DirectXMan12: yes, but we'll have to rev the major version.  I'm curious as to what that situation might be -- you're not using this to detect errors (we have errors for that), so fundamentally, what's the difference here between "a set with no items", and "no set" from the user perspective.

simo5: ok

frozencemetery:
@DirectXMan12 are you happy with the way this is being tested, or do you want something different?

DirectXMan12:
Yeah, I'm not certain there's much more we can do, unless there's a dummy OID that we know will never have any attrs.  This should be good to merge.

PR #151, opened 2018-03-22
Implement remaining GGF extensions

jborean93:
This PR implements the remaining GGF extensions https://github.com/pythongssapi/python-gssapi/issues/51, excluding the `gss_{import,export}_cred` variants as they are not implemented as per the GGF spec in both MIT krb5 and Heimdal. This PR adds in the `gss_set_cred_option` and `gss_set_sec_context_option` functions.

While `gss_set_cred_option` doesn't seem to be a function that was part of the GGF draft, most implementations bundle it together with the GGF `gss_set_sec_context_option` and so I decided to keep it in `ext_ggf`, please let me know if you want to split them up further.

I also have a commented out test for `gss_set_sec_context_option`, it requires `gss-ntlmssp` and unfortunately I was unable to get that working in the docker container tests. Not sure if I needed to do anything extra (apart from installing it from apt/dnf) as it was unable to find the mech when i supplied it as part of the test. For now it is commented out until someone smarter can figure it out.


DirectXMan12:
If `gss_set_cred_option` *could* reasonably not be present while still having the rest present, then it should be in a different file.  In general, the rule is that we never want a canary check (see `setup.py` to pass, but the compile to fail on missing functions or types).

gssapi/raw/ext_ggf.pyx
```diff
@@ -127,3 +141,112 @@ def inquire_sec_context_by_oid(SecurityContext context not None,
...
+        cred_handle (Creds): the Credentials to set, or None to create a new
+            credential.
+        value (bytes): the value to set on the desired aspect of the Credential
+            or None to send GSS_C_EMPTY_BUFFER.
+
+    Returns:
+        Creds: The output credential.
+
+    Raises:
+        GSS_ERROR
```
DirectXMan12: This should be the python type (`GSSError`)

DirectXMan12: (can you fix that up in the rest of these functions, too?  Don't know how that snuck by last time).

gssapi/raw/ext_ggf.pyx
```diff
@@ -127,3 +141,112 @@ def inquire_sec_context_by_oid(SecurityContext context not None,
...
+    set_sec_context_option(desired_aspect, context=None, value=None)
+
+    This method is used to set a value for a specific OID of a
+    :class:`SecurityContext` object. The OID and value to pass in depends on
+    the mech the SecurityContext backs.
+
+    An example of how this can be used would be to reset the NTLM crypto engine
+    used in gss-ntlmssp. The OID that controls this value is
+    '1.3.6.1.4.1.7165.655.1.3' and it takes it a byte value that represents
+    an int32 where 1 reset's the verifier handle and any other int resets the
```
DirectXMan12: s/reset's/resets/

gssapi/tests/test_raw.py
```diff
@@ -847,6 +847,67 @@ def test_inquire_sec_context_by_oid_should_raise_error(self):
...
+        name = gb.import_name(SERVICE_PRINCIPAL,
+                              gb.NameType.kerberos_principal)
+        # GSS_KRB5_CRED_NO_CI_FLAGS_X
+        no_ci_flags_x = gb.OID.from_int_seq("1.2.752.43.13.29")
+        orig_cred = gb.acquire_cred(name).creds
+
+        # nothing much we can test here apart from it doesn't fail and the
+        # id of the return cred is the same as the input one
+        output_cred = gb.set_cred_option(no_ci_flags_x, creds=orig_cred)
+        id(orig_cred).should_be(id(output_cred))
```
DirectXMan12: oof, I don't like testing the `id` like that.  Just drop that line.  It's probably fine to just test that it doesn't error out.

gssapi/tests/test_raw.py
```diff
@@ -847,6 +847,67 @@ def test_inquire_sec_context_by_oid_should_raise_error(self):
...
+                              gb.NameType.kerberos_principal)
+        orig_cred = gb.acquire_cred(name).creds
+
+        # this is a fake OID and shouldn't work at all
+        invalid_oid = gb.OID.from_int_seq("1.2.3.4.5.6.7.8.9")
+        gb.set_cred_option.should_raise(gb.GSSError, invalid_oid, orig_cred,
+                                        b"\x00")
+
+    # TODO: get these tests to detect gss-ntlmssp once it is installed
+    """
```
DirectXMan12: gss-ntlmssp will be installed as part of #150, so we just need to figure out how to detect it, and get that merged into k5test.

gssapi/tests/test_raw.py
```diff
@@ -847,6 +847,67 @@ def test_inquire_sec_context_by_oid_should_raise_error(self):
...
+                                                value=b"\x00" * 4)
+        id(orig_context).should_be(id(out_context))
+    """
+
+    @ktu.gssapi_extension_test('ggf', 'Global Grid Forum')
+    def test_set_sec_context_option_fail(self):
+        # because MIT krb5 doesn't implement any OID's for
+        # gss_set_sec_context_option, we just need to query any OID and it will
+        # raise an exception
+        gb.set_sec_context_option.should_raise(gb.GSSError,
```
DirectXMan12: this test feels fragile.  Is there anything else we can do?

jborean93: Once the NTLMSSP stuff works in the tests we can call that with an invalid value instead.

gssapi/tests/test_raw.py
jborean93: Sounds good.

gssapi/tests/test_raw.py
jborean93: Will do.

gssapi/raw/ext_ggf.pyx
jborean93: Sure, don't know why I wrote them this way.

jborean93:
Thanks for the review @DirectXMan12, let me know what you would like me to do about the `gss_set_sec_context_option` tests considering ntlmssp isn't available right now.

frozencemetery:
It's there on master now.

jborean93:
@frozencemetery, I think there is more that needs to be done. I did that locally before and the tests were still unable to pick up the NTLM mech when creating the credentials/context. I'll try it again as maybe I did something wrong but it seems like something needs to be done in the k5test for it to recognize it.

PR #154, opened 2018-06-25
Use collections.abc on Python 3

tiran:
Signed-off-by: Christian Heimes <cheimes@redhat.com>


DirectXMan12:
Please add a description to the commit and PR description as to why you're making the change.  For example:

> In Python 3.3+ (the earliest version of Python 3 that we support), the the ABCs for collections actually live in `collections.abc`, and should really be referred to from there, even though they're aliased into `collections` for compatibility purposes.

Also, the check should be `if six.PY2` to match the rest of the way we do those checks.

DirectXMan12:
(follow-up -- is the alias being removed from `collections` at a concrete point in the future, or is this just to avoid any potential issues later down the line?)

tiran:
The PR addresses https://bugzilla.redhat.com/show_bug.cgi?id=1594834 . The alias will be removed in 3.8.

DirectXMan12:
ok, that's fine.  Please add it to the commit message and PR description, and I'll merge the PR.

PR #155, opened 2018-09-02
Rename m_ arguments to _attrs in gssapi/mechs.py

cipherboy:
Signed-off-by: Alexander Scheel <ascheel@redhat.com>

This was requested by @DirectXMan12.

Renames the `m_` arguments in `from_attrs` to use `_attrs` instead. Some prefix/suffix is required as `except` is a keyword in Python. However, `m_` was originally chosen as it matched the RFC but isn't descriptive. `_attrs` is more descriptive. As this hasn't been part of a release yet, renaming these arguments should be fine.

### DIFF
--- a/gssapi/mechs.py
+++ b/gssapi/mechs.py
@@ -164,15 +164,16 @@ class Mechanism(roids.OID):
         return cls(m)
 
     @classmethod
-    def from_attrs(cls, m_desired=None, m_except=None, m_critical=None):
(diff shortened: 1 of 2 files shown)


frozencemetery:
@cipherboy Patch no longer works on master (keyword arguments changed, I think).

cipherboy:
@frozencemetery Ah, thanks. :) Updated!

PR #158, opened 2018-10-01
Revert "Expose mechanisms in the high-level API"

frozencemetery:
This reverts commit fe6d5cc634ac09de6634f174db1b598070456a12.

That commit did not pass CI at the time it was added.

[I don't want to merge this commit, but after poking around a bit, accounting for NTLM in the tests and interface is somewhat nontrivial.  @cipherboy @DirectXMan12 Thoughts?]


frozencemetery:
\o/ we didn't have to merge this.

PR #159, opened 2018-10-02
Hopefully fix test issues around high-level mechanism support and indicate/inquire mechs.

cipherboy:
Fixes some issues found by the NTLM mechs:
 - NTLM mechs have empty-but-present SASL names and descriptions
 - Dotted form fallback isn't handled correctly when SASL name is empty
 - Last seen attribute isn't necessarily sufficient for mech attribute inquiry. 

Hopefully these behave a little better. :)

### DIFF
--- a/gssapi/tests/test_high_level.py
+++ b/gssapi/tests/test_high_level.py
@@ -380,18 +380,18 @@ class MechsTestCase(_GSSAPIKerberosTestCase):
     @ktu.gssapi_extension_test('rfc5801', 'RFC 5801: SASL Names')
     def test_sasl_properties(self):
         mechs = gssmechs.Mechanism.all_mechs()
-        encoding = gssutils._get_encoding()
         for mech in mechs:
             s = str(mech)
             s.shouldnt_be_empty()
             s.should_be_a(str)
-            s.should_be(mech._saslname.mech_name.decode(encoding))
 
-            mech.sasl_name.shouldnt_be_empty()
-            mech.sasl_name.should_be_a(six.text_type)
+            # Note that some mechanisms don't have SASL names or SASL
+            # descriptions; in this case, GSSAPI returns empty strings.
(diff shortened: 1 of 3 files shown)


gssapi/tests/test_high_level.py
```diff
@@ -380,18 +380,16 @@ def test_indicate_mechs(self):
...
-        encoding = gssutils._get_encoding()
         for mech in mechs:
             s = str(mech)
             s.shouldnt_be_empty()
             s.should_be_a(str)
-            s.should_be(mech._saslname.mech_name.decode(encoding))
 
-            mech.sasl_name.shouldnt_be_empty()
-            mech.sasl_name.should_be_a(six.text_type)
+            if mech.sasl_name:
```
DirectXMan12: add a comment as to why this is conditional

gssapi/tests/test_raw.py
```diff
@@ -705,7 +707,12 @@ def test_rfc5587(self):
                 display_out.short_desc.should_be_a(bytes)
                 display_out.long_desc.should_be_a(bytes)
 
-        attrs = set([last_attr])
+                if mech_attr not in attrs_dict:
+                    attrs_dict[mech_attr] = 0
+                attrs_dict[mech_attr] += 1
```
DirectXMan12: the comment above needs to reflect the commit message comments to explain what's going on here.
Also, don't you need to ensure that last_mech actually has the attr -- maybe we should record the mechs that have each attr?

DirectXMan12:
Also, please give a better PR title ;-)

gssapi/tests/test_high_level.py
cipherboy: This should be resolved.

DirectXMan12:
Looks reasonable to me.  @frozencemetery can you take a gander, then we can merge?

PR #167, opened 2019-02-19
Fix minor doc typos

lonetwin:


frozencemetery:
Thanks for the patch!  PR to fix CI is https://github.com/pythongssapi/python-gssapi/pull/168 ; will rerun this once it's merged.  It looks like one of our dependencies doesn't work with newer python, so please bear with us as we resolve that.

Issue #50, opened 2015-02-16
Implement RFC 4178 (SPNEGO) related extensions


DirectXMan12:
RFC 4178 contains two methods for use with SPENGO: 
- gss_set_neg_mechs
- gss_get_neg_mechs (not implemented in MIT krb5)

These are useful to constrain the mechanisms that SPENGO can use

optiz0r:
The code in #176 seems to work for me, it would be useful if this could be published as a new release :)

frozencemetery:
Thanks for the feedback @optiz0r!  I'm waiting on Windows support (#177) which should merge soon, and then I'll cut a release.

(Closing ticket because it should have been closed with the merge of #176.)

PR #176, opened 2019-06-04
Implemented RFC 4178 extension 

tengcm2015:
reference issue #50


gssapi/raw/ext_rfc4178.pyx
```diff
@@ -0,0 +1,54 @@
...
+        OM_uint32 *minor_status,
+        gss_cred_id_t cred_handle,
+        const gss_OID_set mechs) nogil
+
+
+def set_neg_mechs(Creds cred_handle not None, mechs not None):
+    """
+    set_neg_mechs(Creds cred_handle not None, mechs=None)
+
+    This allows callers to specify the set of security mechanisms that
```
frozencemetery: We don't typically use the RFC text for descriptions - it tends to be overly lengthy and include a lot of details we don't care about.  See if you can make a shorter description in your own words.

gssapi/raw/ext_rfc4178.pyx
```diff
@@ -0,0 +1,54 @@
...
+    may be negotiated with the credential identified by cred_handle.
+    This call is intended to support specialized callers who need to
+    restrict the set of negotiable security mechanisms from the set of
+    all security mechanisms available to the caller (based on available
+    credentials). Note that if more than one mechanism is specified in
+    mech_set, the order in which those mechanisms are specified implies a
+    relative preference.
+
+    Args:
+        cred_handle (Creds): the Credentials to query
```
frozencemetery: Technically not a query - that would be more true of `gss_get_neg_mechs()` (which no one implements).

gssapi/raw/ext_rfc4178.pyx
```diff
@@ -0,0 +1,54 @@
...
+        mechs ([MechType]): the desired set of mechanisms that
```
frozencemetery: Maybe shorten this too

gssapi/raw/ext_rfc4178.pyx
```diff
@@ -0,0 +1,54 @@
...
+            may be negotiated with the credentials
+    Returns:
+        None (If the operation completed successfully)
```
frozencemetery: The parenthetical is unnecessary.

gssapi/raw/ext_rfc4178.pyx
```diff
@@ -0,0 +1,54 @@
...
+        maj_stat = gss_set_neg_mechs(&min_stat, cred_handle.raw_creds,
+                                     desired_mechs)
+
+    cdef OM_uint32 tmp_min_stat
+    gss_release_oid_set(&tmp_min_stat, &desired_mechs)
+
+    if maj_stat == GSS_S_COMPLETE:
+        return None
+    else:
+        print("error: %d:%d", maj_stat, min_stat)
```
frozencemetery: Leftover debugging line?

gssapi/tests/test_raw.py
```diff
@@ -772,6 +772,34 @@ def test_sasl_names(self):
             cmp_mech.shouldnt_be_none()
             cmp_mech.should_be(mech)
 
+    @ktu.gssapi_extension_test('rfc4178', 'Negotiation Mechanism')
+    def test_set_neg_mechs(self):
```
frozencemetery: While I think we want to enforce mech_set not being None (even though 4178 doesn't require it), I would like to see us test mech selection.  We should have the NTLM mechanism around - maybe something using that?  (See test_set_sec_context_option and test_set_sec_context_option_fail.)  You'll need to install gssntlmssp to check this locally.

frozencemetery:
Extension detection doesn't seem to have worked on Heimdal because our Debian's Heimdal is too old.  (It was added this year in 735039dbdc3aa58d06afdefd214efe3f5e421244.)  However, it'll work whenever that updates - they just put it in gssapi.h.

@DirectXMan12 PTAL

DirectXMan12:
my context for this is mostly lost to time.  Which part did you want me to take a look at?  The extension negotiation?

frozencemetery:
I was just hoping for code review :)

frozencemetery:
@tengcm2015 Hey, this is still waiting on you.  Any updates?

tengcm2015:
@frozencemetery I fixed what is mentioned in the code review, but have no idea why the test is failing only on centOS

frozencemetery:
Does it work for you locally on centos?

tengcm2015:
Well, I guess the problem was that centOS had no ntmlssp, but since I created a credential with all_mech, it just created a credential with some other mech rather than skipping the test.

Issue #32, opened 2014-12-23
Support Windows?


DirectXMan12:
We may wish to consider supporting Windows at some point

filmor:
What needs to be done to get there?

DirectXMan12:
1. Compilation needs to be tested in Windows with MIT Kerberos for Windows on MinGW.
2. The conditional compilation logic needs to be adjusted to work for Windows
3. We need a way to build wheels for windows (optional, but useful for anyone who actually wants to use this on Windows).

frozencemetery:
Other "nice to have" - some way to run our test suite, since I don't think k5test will work there.

ibigpapa:
So using the MIT KfW.  I see the source is available as well as precompiled sources could we just wrap these to make it work?

DirectXMan12:
yeah, it really comes down to finding time and a windows box to test on (plus setting up CI to make sure we don't break anything), plus tweaking `setup.py` to work properly on Windows.

ibigpapa:
I'm working on trying to hack this in myself let you know how i do.

DirectXMan12:
cool.  Feel free to ping us on #python-gssapi on freenode if you have any questions best answered in a real-time manner (I'm usually on from 10:00-17:00EST)

frozencemetery:
2a2b47b6f35a0058eb8c33e23e7630f6ecdbf5d1 adds support for conditional compilation.  Still missing are building wheels and running buildjob CI.  Much like with Heimdal, a stretch goal is to port k5test so we can run the test suite there.

aiudirog:
I've been working on Windows support the past couple of days to get my company to move off of requests-kerberos (I had previously patched ccs-kerberos to work on Windows) and got everything to compile properly under VS Build Tools, MinGW, and MSYS. Unfortunately I won't be able to test until I'm back at work on Monday since k5test isn't ported yet. [My current progress can be found here.](https://github.com/aiudirog/python-gssapi/tree/aiudirog-win32-support) 

Changes:
- I created a predefined macro (`__MSYS__`) to allow the header files to detect MinGW under MSYS so that when it isn't under it, they can target MIT Kerberos for Windows instead (which also seems to have better extension support). 
- To get it compile under VS:
[...]

PR #177, opened 2019-06-15
Windows Support

aiudirog:
Pull for #32. Currently working on Travis building, no wheels yet. I will squash all of the commits once it's ready.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,6 +9,8 @@ import platform
 import re
 import sys
 import os
+import shutil
+import shlex
 
 
 SKIP_CYTHON_FILE = '__dont_use_cython__.txt'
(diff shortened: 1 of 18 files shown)


aiudirog:
Failing for flake8 checks. I'll take care of those in a bit.

aiudirog:
Need to find a way to install kerberos for windows. msiexec is failing.

aiudirog:
I've squashed all the commits and it's ready for review.

frozencemetery:
Great!  Some initial thoughts (please bear with me, I know very little about Windows):

- I'm hesitant to add more python2 support at this juncture, given the [looming deprecation](https://pythonclock.org/).  Do you have an argument in favor of a python2 builder?

- This looks like it's adding support for two different ways to build on Windows; do I have that right?  Probably there should be some information in the README about how to do that.

- The `<char *>` stuff is a bit odd; do you have more information on why that's needed?  It seems like it's a problem that maybe we should solve at a different layer.

- How "locked" are Windows python versions?  Could we just depend on "3" and get the latest one?  (Is there a reason not to just take the latest 3?)

@DirectXMan12 this touches some of your magic deployment goo, so I do want your signoff before merging.  (Also with your username, how could I not.)

aiudirog:
- I would love to drop Python 2 support as it requires extra steps and forced me to go through two more debugging commits. I implemented it because the other OSs had it.

- It can be built 2 ways with 3 compilers. The first is against MIT Kerberos on both the Visual Studio C++ compiler and MinGW. The second is under MSYS against heimdal using MinGW. While MSYS has many less gss features, support was mostly there before I started but conflicted with the regular MinGW support so I cleaned it up. The wheels are built using VS C++, but it is quite easy to change to MinGW (in fact I was forced to for Py2 as it's VS compiler isn't supported anymore).

- gss_OID_desc.elements is declared as a void* in the C headers but always treated as a char*. The VS C++ compiler refuses to do pointer arithmetic on void* whereas most compilers simply ignore it. Changing the type in the pxd doesn't work since they don't actually get compiled. If we only allow building on MinGW, the char* casts aren't needed but if people want to build themselves it can get finicky as the VS one is preferred as CPython is built with it.

- Wheels are locked to a major/minor version for C-extensions which is why I chose the latest versions for 3.6 and 3.7. I think we could probably change it to look up the 2 or 3 latest versions and build against those.

frozencemetery:
I see, the arithmetic occurs because of the way cython handles `buf[:len]` internally.  (You can easily do this without needing pointer arithemtic, and I guess they chose not to.)  That's unfortunate, but I think you're right that the `char *` cast is the right approach.

Will wait for @DirectXMan12's thoughts on the rest.

aiudirog:
I can also look into opening a pull request with Cython to eliminate the zeros. I'm not thrilled about the char* casts having to be maintained when new extensions have to be added.

aiudirog:
I'm just realizing a made a mistake on my explanation for the char* casts. I was on mobile while refreshing myself on why they were needed and had the wrong struct. 

It's `gss_buffer_desc.value` that I had to cast, not `gss_OID_desc.elements`. The later does need them but has always had them. I was forced to add them to `gss_buffer_desc.value` because even though it is declared in cython_types.pxd as a char*, gssapi.h on Windows, Ubuntu, and Arch all declare it as a void*. Since GCC is used most of the time, this may have gone unnoticed.

frozencemetery:
Interesting.  I appreciate you looking into it, but the reality is that we have to support older cython versions anyway, so it wouldn't help much.  I still think the casts are the right way to go.

setup.py
```diff
@@ -42,47 +43,98 @@ def _get_output(*args, **kwargs):
     get_output = _get_output
 
 # get the compile and link args
-link_args = os.environ.get('GSSAPI_LINKER_ARGS', None)
-compile_args = os.environ.get('GSSAPI_COMPILER_ARGS', None)
+link_args, compile_args = [
+    os.environ[e].split() if e in os.environ else None
```
DirectXMan12: this probably needs to be shsplit if we're going to do it like this

setup.py
```diff
@@ -42,47 +43,98 @@ def _get_output(*args, **kwargs):
...
+        link_args = (
+            ['-L%s' % _libs]
+            + ['-l%s' % os.path.splitext(lib)[0] for lib in os.listdir(_libs)]
+        )
     elif os.environ.get('MINGW_PREFIX'):
-        link_args = '-lgss'
+        link_args = ['-lgss']
     else:
-        link_args = get_output('krb5-config --libs gssapi')
+        link_args = get_output('krb5-config --libs gssapi').split()
```
DirectXMan12: ditto here for shsplit

setup.py
```diff
@@ -42,47 +43,98 @@ def _get_output(*args, **kwargs):
...
+        ]
     elif os.environ.get('MINGW_PREFIX'):
-        compile_args = '-fPIC'
+        compile_args = ['-fPIC']
     else:
-        compile_args = get_output('krb5-config --cflags gssapi')
-
-link_args = link_args.split()
-compile_args = compile_args.split()
+        compile_args = get_output('krb5-config --cflags gssapi').split()
```
DirectXMan12: ditto here

DirectXMan12:
hey, sorry for all the delay, been swamped with other stuff :-(

aiudirog:
No problem for the delay. 

I took care of those three splits. (Should have done that myself originally, haha)

Overall improvements to setup.py might want to be its own PR. One improvement I had in mind was  moving the shared kwargs to `Extension` into a dictionary or having a dedicated function to avoid repeating them.

frozencemetery:
@aiudirog If you're willing to make other improvements, feel free to tack them onto this PR as a separate commit - I think that should be fine.  (And if you're not, just let us know :) )

aiudirog:
No problem, I started by abstracting out the `Extension` initialization and dropping the Py26 support for `commands.getoutput`.

I'm not quite sure yet of the best way to clean up the rest. Not too much is actually repeated or easily abstracted into useful functions, so maybe the best way would be to have one function for each environment that configures it completely? However that would split up the configuration stages so it wouldn't flow as well top to bottom. Maybe sectioning off each block with clearer/more thorough commenting is the way to go? 

Any thoughts?

setup.py
```diff
@@ -28,20 +29,18 @@
...
         SOURCE_EXT = 'c'
 
-get_output = None
 
-try:
-    import commands
-    get_output = commands.getoutput
-except ImportError:
-    import subprocess
+try:  # TODO: Remove this when Py27 is dropped
```
frozencemetery: Move to just below the imports.

aiudirog: After thinking about it again, I decided to just use OSError instead of FileNotFoundError as it's close enough and works without the hack.

aiudirog:
I decided to drop the 2.6 support because it wasn't in the classifiers anyways and you had mentioned dropping 2.7 support earlier. 

It looks like there are 4 distinct build paths (Windows, OSX > 10.7, MSYS, & Linux/OSX < 10.7), then the varying support like Heimdal which affects MSYS and Linux. I think I'll give it a try and put it in a separate branch. I'll ping you when it's ready and if we don't like it, we can just throw it away.

aiudirog:
@frozencemetery [Let me know what you think.](https://github.com/aiudirog/python-gssapi/commit/cacf6a645149a699c4d288283df081ad47fe0521) After testing a few different solutions, I found that having a default config that does most of the heavy lifting and then subclassing it to customize for each platform feels the cleanest. Tested so far on Arch, Windows, and MSYS.

PR #179, opened 2019-07-17
Work around PyPI not working on Windows in Travis

aiudirog:
@frozencemetery When you get a chance, see if this fixes the wheel deployment for now. I just need you to replace the template credentials with the encrypted ones.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -144,22 +144,31 @@ jobs:
     stage: deploy
     os: windows
     script: skip
-    env: PYTHON="2" PYENV="2.7.16" EXTRA_BUILDEXT="--compiler=mingw32"
+    env:
+      - PYTHON="2"
+      - PYENV="2.7.16"
+      - EXTRA_BUILDEXT="--compiler=mingw32"
+      - secure: "PIP_USER=USER"  # <- Replace and encrypt
+      - secure: "PIP_PASSWORD=PASSWORD"  # <- Replace and encrypt
     before_deploy:
       - ./.travis/before-deploy-windows-wheels.sh
     deploy:
-      - provider: pypi
-        user:
-          secure: "jUAMucBq+9xH8x9u0I0LOwrs3Zb++KN7FwIIwz2CyAt/+TyyrJzeGJaV+dTiJ1OqcUIFqQG6jopzpnAe4biL1O68PEwz9BphKetFLpLHiFNm/n67LYno6NFonWmxndIy99pOP6NZu29nzSNeYq/KgEHo/5OkqEGOxk//lh7X/OY="
-        password:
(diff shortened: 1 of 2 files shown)


aiudirog:
Adding tag back to issue. #32

frozencemetery:
Hi @aiudirog, thanks for the PR.  I'm doing some test builds with it - deploys are very slow so it will take a while.

One important change I've had to make is around environment variables: Windows build log obfuscation doesn't seem to work (per [here](https://docs.travis-ci.com/user/best-practices-security)) so this will take several attempts.

frozencemetery:
Fun fact: pip reads all env vars of the form PIP_NAME=VALUE as if they were passed as arguments `pip --name=value`.

aiudirog:
Looks like they still have a lot to improve. One other thing that might work is building the wheels after testing and dumping them somewhere to be picked up by the Linux before-deploy.sh. 
--------
It does the same thing with pip.conf. I remember spending a while trying to find what options could be used in it and it took longer than I care to admit to realize it was just all command line options.

frozencemetery:
@aiudirog Do you have ideas on how to do that, or should we just give up and hand-upload Windows wheels for now?

aiudirog:
Travis seems to recommend storing artifacts which need to persist between build stages in an S3 bucket. The free tier would probably work perfectly for it.

aiudirog:
@frozencemetery This sounds like a hack but it might work: we use [s3 deployment](https://docs.travis-ci.com/user/deployment/s3/) on the individual Windows build steps to upload the wheels and then add a 4th stage which runs in a Linux container that just pulls the wheels back and uploads them to PyPi. Alternatively, we move the Windows deploy stages above the Linux one so that they run first and then have the Linux one grab and upload the wheels without adding the extra dedicated stage. 

If that sounds too hacky, at first glance AppVeyor appears to be mostly cmd/PowerShell so more than expected might have to be implemented to get it running. Hand uploading might actually be the most painless way to go until this issue is resolved. I do all of my Windows work in a dedicated VM so I could build the wheels and ship them over to you whenever there is a release or we could work something else out.

Thoughts?

frozencemetery:
So I've got a branch that successfully uploads to pypi - you can see the wheels are successfully present on [the download page](https://pypi.org/project/gssapi/0.0.2/#files).  The problem is that this bypasses Travis's deploy logic entirely - I need to turn it off.  I think this is easy, but Travis is having stability issues right now, so this will probably have to wait another day.

When that's done, I'll squash it all and cut 1.6.1.  After that, I want to look into paring down our test matrix - this takes forever to run, and having this many discrete travis items exposes us to many infrastructure failures Travis has.  Probably we can eliminate the verify stage, and then in January we can drop out all the python2 builders.

frozencemetery:
(Superseded by #180.)

PR #184, opened 2019-08-20
Update cred store docs and restore behaviour of raising exception on a None value

frozencemetery:


gssapi/raw/ext_cred_store.pyx
```diff
@@ -83,8 +83,7 @@ cdef gss_key_value_set_desc* c_create_key_value_set(dict values) except NULL:
 
     for (i, (k, v)) in enumerate(values.items()):
         res.elements[i].key = k
-        if v:
```
simo5: should we raise a specific exception here ?
Or just let python throw up whatever ?

frozencemetery: It throws up `TypeError: expected bytes, NoneType found` which I think is correct - we want a (byte)string there.  Do you think there's something better?

simo5: I guess that's a good exception, I just asn't sure what it would throw.

Issue #189, opened 2020-02-03
Python 3.8 on Windows


ofek:
Hello! Are there plans to ship binary wheels for Python 3.8 on Windows soon? Many of us would really appreciate it.

Thanks 😄 

cc @frozencemetery

Issue #181, opened 2019-08-09
Reduce runtime of CI


frozencemetery:
### What went wrong?

It takes a really long time to run through the full CI matrix - and even longer to do a full deploy.  

### More information

Our test suite itself is fast, and our build times aren't too shabby.  We waste most time in two places: pulling down data from mirrors (e.g., pypi, distro mirrors, docker hub), and waiting on Travis to allocate us machines.

### Suggested fix

- [ ] Drop the python2 builders in January, which reduces number of machine requests
- [ ] Combine the first two stages so that we have fewer total machine requests
- [ ] When CentOS 8 becomes available, replace the el7 builder with it (ideally this happens before January)

PR #190, opened 2020-02-05
Release updates for python-3.8 and removing python2

frozencemetery:
Resolves: #189
Resolves: #181


frozencemetery:
Windows contributors (@aiudirog @totaam @ofek n'at), the obvious change for removing python2 drops our mingw32 build/release pipeline.  That probably shouldn't go away, right?  What should the Windows matrix look like ideally?

frozencemetery:
Have given up on updating the MacOS builder and opened https://github.com/pythongssapi/python-gssapi/issues/191 for that.

Will merge and tag once we know what the Windows should look like.

ofek:
Thanks!

aiudirog:
I'll look over all of the Windows stuff when I have some time tomorrow evening.

aiudirog:
For removing Python 2.7 on Windows, it looks like the only remaining cleanup is to remove the [extra dependency installation](https://github.com/pythongssapi/python-gssapi/blob/51c83f854c593a52a829ce1a3097c3c9cf4a08bf/.travis/lib-setup.sh#L87). 

On another note, I would personally restore 3.6 wheel deployment on Windows since its EOL isn't until the end of 2021.

We may also want to bump all 3.7 testing and deployment up to 3.7.5 where available, but that's not really a big deal.

frozencemetery:
Thanks for your patience.  I've updated with @aiudirog's suggestions.

Issue #193, opened 2020-03-22
ImportError: DLL load failed: The specified module could not be found on Windows


atodorov:
### What went wrong?

Trying to use gssapi on Windows seems to fail:

```
Traceback (most recent call last):
  File "setup.py", line 4, in <module>
    from tcms_api.xmlrpc import __version__
  File "C:\projects\tcms-api\tcms_api\__init__.py", line 72, in <module>
    from .xmlrpc import TCMSXmlrpc, TCMSKerbXmlrpc
  File "C:\projects\tcms-api\tcms_api\xmlrpc.py", line 11, in <module>
    import gssapi
  File "C:\Python36-x64\lib\site-packages\gssapi\__init__.py", line 29, in <module>
    from gssapi.raw.types import NameType, RequirementFlag, AddressType  # noqa
  File "C:\Python36-x64\lib\site-packages\gssapi\raw\__init__.py", line 50, in <module>
    from gssapi.raw.creds import *  # noqa
ImportError: DLL load failed: The specified module could not be found.
```

### How do we reproduce?

`import gssapi` seems to be enough

### Component versions (python-gssapi, Kerberos, OS / distro, etc.)

gssapi-1.6.2, pip install - downloads the wheel package, Windows. 

Logs at:
https://ci.appveyor.com/project/atodorov/tcms-api/builds/31627921/job/3tbk1ike6er25tmd

Feels like something is missing but I don't know what. Works perfectly fine on Linux.

Update: looks like #188 but I couldn't figure out what the solution was.

atodorov:
More info: 

I tried this manually on a Windows Server 2019 system:

1) Install Pythin 3.8 64bit
2) Install MIT Kerberos for Windows, 64bit, Complete installation
3) Create a venv and pip install gssapi

The wheel package installs but import still errors out in the same way. 

How are the supplied wheel packages built and what libraries do they require ?

frozencemetery:
#188 was solved by installing krb5.  Can you confirm you did that?  The latest KfW should have cred_store support.

CC @aiudirog

aiudirog:
The DLLs for K4W are probably what's missing. I really wish DLL errors would actually tell you which one haha. If you haven't already, install K4W and then make sure it is in your PATH. It should be enough to just add `C:\Program Files\MIT\Kerberos\bin` if it isn't.

[...]

PR #197, opened 2020-04-03
Fix Loading DLL on Windows Python >=3.8

aiudirog:
As discussed in #193, Python 3.8 no longer searches the PATH for DLLs on Windows and instead looks in standard locations. I've updated the Windows compatibility handling to account for this and use the new `os.add_dll_directory()` when possible.

I also modified the tests to at least make sure the library can be imported on systems that we can't test on.  This would have failed the 3.8 build last release and caught the issue earlier.

### DIFF
--- a/.travis/build.sh
+++ b/.travis/build.sh
@@ -30,19 +30,15 @@ fi
 python setup.py build_ext --inplace $EXTRA_BUILDEXT
 BUILD_RES=$?
 
-if [ x"$KRB5_VER" = "xheimdal" ]; then
-    # heimdal can't run the tests yet, so just exit
-    exit $BUILD_RES
-fi
-
-if [ "$TRAVIS_OS_NAME" == "windows" ]; then
-    # Windows can't run tests yet, so just exit
-    exit $BUILD_RES
-fi
-
 if [ $BUILD_RES -ne 0 ]; then
     # if the build failed, don't run the tests
     exit $BUILD_RES
+fi
+
+if [ x"$KRB5_VER" = "xheimdal" ] || [ "$TRAVIS_OS_NAME" = "windows" ]; then
+    # heimdal/Windows can't run the tests yet, so just make sure it imports and exit
+    python -c "import gssapi"
+    exit $?
 fi
 
 python setup.py nosetests --verbosity=3
--- a/gssapi/_win_config.py
+++ b/gssapi/_win_config.py
@@ -7,6 +7,7 @@ down the import tree.
 """
 
 import os
+import shutil
 import ctypes
 
 #: Path to normal KfW installed bin folder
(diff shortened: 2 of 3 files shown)


aiudirog:
Annnnnd the testing requirements aren't installed on the ones we don't run tests on (－‸ლ)

gssapi/_win_config.py
```diff
@@ -41,14 +42,31 @@ def error_not_found():
...
+
+    # Check if kinit is in the PATH which should lead us to the bin folder
+    kinit_path = shutil.which('kinit')  # MIT exclusive binary
+    if kinit_path:  # Non-standard install location
+        try:  # Most likely >=3.8, otherwise it would have been found already
+            os.add_dll_directory(os.path.dirname(kinit_path))
+        except AttributeError:  # <3.8, corrupted installation?
+            pass
+        else:
+            if kfw_available():
```
frozencemetery, line 67: Collapse this into `elif`.

gssapi/_win_config.py
```diff
@@ -41,14 +42,31 @@ def error_not_found():
...
-        if k4w_in_path():
+        try:  # to use Python 3.8's DLL handling
+            os.add_dll_directory(KFW_BIN)
+        except AttributeError:  # <3.8, use PATH
+            os.environ['PATH'] += os.pathsep + KFW_BIN
+        if kfw_available():
             return
+
+    # Check if kinit is in the PATH which should lead us to the bin folder
+    kinit_path = shutil.which('kinit')  # MIT exclusive binary
```
frozencemetery, line 60: I'm not sure I understand this comment.  Is the intent to check for a binary that can only come from KfW?  If so, perhaps this should read "KfW-provided binary" or something?

But I'm also probably in the minority of first thinking about Heimdal whenever something says "MIT exclusive", though :)

.travis/build.sh
```diff
@@ -30,19 +30,21 @@ fi
...
+if [ x"$KRB5_VER" = "xheimdal" ]; then
+    # heimdal can't run the tests yet, so just make sure it imports and exit
+    python -c "import gssapi"
+    exit $?
 fi
 
-if [ $BUILD_RES -ne 0 ]; then
-    # if the build failed, don't run the tests
-    exit $BUILD_RES
+if [ "$TRAVIS_OS_NAME" == "windows" ]; then
```
frozencemetery, line 44: Same check in both cases - can you fold it into a single conditional?

aiudirog: Sure, it was two separate checks before, so I just left it as such

gssapi/_win_config.py
aiudirog: Yes and I'll make the change

gssapi/_win_config.py
aiudirog: That's a try block, not an if statement:P At first I was like, try blocks support elif??? but they don't:(

aiudirog:
I figured you would squash the commits like you did for the previous PR so the commit order wouldn't matter. I'll push with new commits in that order.

gssapi/_win_config.py
frozencemetery: Welp, TIL.  Weird inconsistency :/

Issue #198, opened 2020-04-08
setup.py should have python_requires


jborean93:
### What went wrong?
The new 1.6.4 release dropped support for Python 2.7 and 3.5 which is fine but when a user goes to `pip install gssapi` on those hosts it will install 1.6.4 when 1.6.2 should really be the one to install.

I'm not sure if we can actually do anything for this now a release is out but it would be good to avoid this in the future when a Python version is dropped. IIRC to do this you just need to add `python_requires` to setup.py which is what Ansible is doing https://github.com/ansible/ansible/blob/148e83f832aafa27bb49e98667bdcf72f1fd0f84/setup.py#L277. For this project it should be `python_requires=">=3.6.*"`

### How do we reproduce?

`pip install gssapi` on a Python 2.7 or 3.5. The install works fine but when you go to use the library 3.6 specific syntax will cause a failure. Ideally a pip install on those version should select the latest version that is supported for that Python version.

### Component versions (python-gssapi, Kerberos, OS / distro, etc.)

Install through pip on Python 2.7 and/or 3.5.

briantist:
Thanks for submitting this; this burned me hard yesterday.

frozencemetery:
Ah, will fix in a moment.  Thanks for pointing this out.

We aren't interested in python 2 - it's done upstream and we haven't supported it since January.  However, 3.5 is still supported upstream, so I'll adjust this.

frozencemetery:
I've unpushed the 1.6.4 release from pypi and will cut a 1.6.5 with the changes from #199.  This should fix the problems observed, I think.

PR #199, opened 2020-04-08
Add python_requires since upstream still supports 3.5

frozencemetery:
Resolves: #198

@jborean93 PTAL

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -304,6 +304,7 @@ setup(
     long_description=long_desc,
     license='LICENSE.txt',
     url="https://github.com/pythongssapi/python-gssapi",
+    python_requires=">=3.6.*",
     classifiers=[
         'Development Status :: 5 - Production/Stable',
         'Programming Language :: Python',


jborean93:
Just wanted to check that when you build the package the setuptools version is at `>=24.2.0` as that is what is required for the metadata to be read and generated properly https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires.

Also you state that upstream still supports 3.5 but you've used an fstring with https://github.com/pythongssapi/python-gssapi/commit/5e91aa07271af305b78c9ac45d741f83035ea459#diff-d7c47bf8b55e340087c202452e7c7a8eR64 which is a Python 3.6 only change. The setup.py contents state 3.6+ as well so I'm just a bit confused by the title here.

frozencemetery:
Ah, good question.  "Upstream" there referred to "upstream python".  I'll fix the commit message.

frozencemetery:
To your other question: we build and upload the packages from fedora:latest (currently 31) so we should be good there, and the Windows builds are done from pip.

jborean93:
Awesome thanks for the clarification here and the quick turnaround. Much appreciated.

PR #204, opened 2020-05-08
Migrate CI and automation to GitHub actions

frozencemetery:
(Supersedes #201)

I took the work @atodorov had done and migrated the parts that require additional repo access as well.  A couple things to note:

- The official pypi upload action doesn't work on Windows.  I'm told this is on purpose, but their suggested workflow is significantly more complex, so I'm keeping the old approach instead.
- I changed the scripts directory to "ci" instead of ".ci" because it was easier to type and it doesn't need to be hidden.  I'd change ".github" too if I could :)
- Windows python versions are allowed to float in the last digit now.  I think this is a positive change, though it's not required.
- Travis button goes away because we stop using Travis
- Installing packages on Windows is extremely fraught


frozencemetery:
I have tested the release workflow and know it works in the way I'd expect.  This PR will test the PR workflow :)

PR #205, opened 2020-05-11
Build & push wheel packages on Linux. Fixes #200

### LINKED ISSUE #200: Ship wheel packages for Linux ?
### What went wrong?

How do maintainers feel about providing wheel packages for Linux? Maybe 64bit only, same Python versions for which you build the Windows wheel packages ?

atodorov:
@frozencemetery please review

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -63,6 +63,50 @@
             ],
         },
 
+        "wheels-linux": {
+            "runs-on": "ubuntu-latest",
+            "strategy": {
+                "matrix": {
+                    "name": ["linux-wheel-3.8", "linux-wheel-3.7",
+                             "linux-wheel-3.6"],
+                    "include": [
+                        {
+                            "name": "linux-wheel-3.8",
+                            "pyenv": "3.8",
+                        },
+                        {
+                            "name": "linux-wheel-3.7",
+                            "pyenv": "3.7",
+                        },
+                        {
+                            "name": "linux-wheel-3.6",
+                            "pyenv": "3.6",
+                        },
+                    ],
+                },
+            },
+            "steps": [
+                {
+                    "name": "Check out code",
+                    "uses": "actions/checkout@v2",
+                },
+                {
+                    "name": "Install the right python",
+                    "uses": "actions/setup-python@v2",
+                    "with": { "python-version": "${{ matrix.pyenv }}" },
+                },
+                {
+                    "name": "Create and upload Linux wheel",
+                    "shell": "bash",
+                    "run": "./ci/release-wheel.sh",
+                    "env": {
+                        "TWINE_USER": "rharwood",
+                        "TWINE_PASSWORD": "${{ secrets.pypi_password }}",
+                    },
+                },
+            ],
+        },
+
         "release-windows": {
             "runs-on": "windows-latest",
             "strategy": {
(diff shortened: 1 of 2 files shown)


ci/before-deploy.sh
```diff
@@ -50,3 +55,7 @@ PKG_NAME_VER="python-gssapi-${PYTHON_GSSAPI_VERSION}"
 
 tar -czvf ./tag_build/${PKG_NAME_VER}.tar.gz --exclude='dist' --exclude='tag_build' --exclude='.git' --exclude='travis_docs_build' --exclude='.git' --transform "s,^\.,${PKG_NAME_VER}," .
 sha512sum --binary ./tag_build/${PKG_NAME_VER}.tar.gz > ./tag_build/${PKG_NAME_VER}.sha512sum
+
+# Rename and checksum the wheel
+mv dist/*.whl dist/"${PKG_NAME_VER}.whl"
```
frozencemetery, line 60: This produces a file named: python-gssapi-1.6.5.whl.  Looking at [PEP 427](https://www.python.org/dev/peps/pep-0427/#file-name-convention), I'm not convinced this is right.

More generally, we have C code.  I believe this means we need to do for Linux what we do for Windows and build three times.  I'm not sure if it's best to add two or three more jobs for that.

frozencemetery: If you would prefer me to just do this myself instead, let me know :)

atodorov: Right, the name is incorrect, should be underscores instead of dashes. Which begs the question why don't we use the correct name in setup.py and sdist/bdist will produce the right one?

atodorov: > More generally, we have C code. I believe this means we need to do for Linux what we do for Windows and build three times.

I have not explored that in details, I just saw a `python setup.py bdist_wheel` and gave it a go. What do you mean by "build 3 times" ?

frozencemetery: Well, you did the correct build, but then you renamed it :)

Take a look at the Windows wheels here: https://pypi.org/project/gssapi/#files

There are three of them - and they're specific to python version, which is what I was trying to highlight by linking the PEP.  If you look at the Windows wheel building (in release.yml), there are in fact three separate jobs - one for each python version we build for.

I'm not aware of a way to make one build work for multiple python versions, so my inclination would be to do one for each of the three we support, as we do for Windows.  Does that make sense?

atodorov: > Well, you did the correct build, but then you renamed it :)

I got confused by the rename in the windows release script. 

For the multiple python versions: they can be specified in the job matrix but I think before-deploy.sh runs inside a docker container so not sure how to specify different python version in there.

atodorov: Would a `dnf install pythonX.Y` inside the Fedora container suffice ?

frozencemetery:
No, it won't work because Fedora only packages one version of python3 at a time in released distros.  We should be fine doing those builds outside the container though - see what I pushed.

The rename just turns python-gssapi into gssapi.  On my local machine, it's in fact a no-op - the files are created as, e.g., "gssapi-1.6.5-cp38-cp38-linux_x86_64.whl".  I expect it was a workaround for an old version of something and could go away at this point.

atodorov:
So this has been resolved by your latest commit then?

frozencemetery:
(Closing - move discussion to #200)

PR #207, opened 2020-05-20
Test manylinux2014 wheels

frozencemetery:
Resolves: #200

For testing purposes, wheels were uplaoded to TestPyPI: https://test.pypi.org/project/gssapi/1.6.5/#files

Thanks to @tiran for assistance!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,7 +2,7 @@
     "name": "Release",
     "on": { "push": { "tags": "v*" }},
     "jobs": {
-        "release-linux": {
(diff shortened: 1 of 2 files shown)


frozencemetery:
@tiran @simo5 Opinions on merging this?  Something I should do different?

tiran:
I'm still worried that the binary wheels are going to break stuff. I don't have time to check how binary wheel python-gssapi with its own KRB5/GSSAPI libraries interact with python-ldap builds that use the system KRB5/GSSAPI libraries.

Issue #208, opened 2020-06-01
basic-tutorial.md -- name should not equal server_name


rebeccc:
### What went wrong?
On line 85 of basic-tutorial.md, the call to Credentials fails because it is server_name and not server_hostbased_name

### How do we reproduce?
Based on the tutorial, you set server_name and server_hostbased_name as follows:
```
server_hostbased_name = gssapi.Name('HTTP@' + FQDN, name_type=gssapi.NameType.hostbased_service)
server_name = gssapi.Name('HTTP/sross@')
```
and then make a call to Credentials using server_name rather than server_hostbased_name, which causes it to fail.
```
server_creds = gssapi.Credentials(usage='accept', name=server_name)
```

rebeccc:
There are a couple of other places in the tutorial that uses server_name instead of server_hostbased_name. I've made a list of the ones I realized wouldn't work. 

this one is just in a sequence of checks and doesn't fail, but will show up false.
line 103:
```
server_creds.name == server_name
```
line 128:
```
client_ctx = gssapi.SecurityContext(name=server_name, usage='initiate')
```

frozencemetery:
Fixed by #209

PR #209, opened 2020-06-02
changing server_name to server_hostbased_name

rebeccc:
fixes issue #208

### DIFF
--- a/docs/source/basic-tutorial.md
+++ b/docs/source/basic-tutorial.md
@@ -43,12 +43,12 @@ Suppose we wanted to refer to an HTTP server on the current host.
 We could refer to it as a *host-based service*, or in the default
 mechanism form (in this case, for krb5):
 
-    >>> server_hostbased_name = gssapi.Name('HTTP@' + FQDN, name_type=gssapi.NameType.hostbased_service)
+    >>> server_hostbased_name = gssapi.Name(f"HTTP@{FQDN}", name_type=gssapi.NameType.hostbased_service)
     >>> server_hostbased_name
-    Name(b'HTTP@sross', <OID 1.2.840.113554.1.2.1.4>)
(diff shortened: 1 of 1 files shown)


rebeccc:
@frozencemetery Changing ` server_hostbased_name ` to ` server_name` makes better sense to me, too. 

I'm not really sure what to call HTTP/sross@ because I'm not exactly sure what it was trying to show in the tutorial in the first place. I can tell that the tutorial was trying to show the difference between a Name with and without the `name_type = gssapi.NameType.hostbased_service`, and that they can successfully be compared after canonicalizing them. Should HTTP/sross@ be changed to `'HTTP@' + FQDN`?

frozencemetery:
That's a good point.  After rereading and thinking about this some more, I think Solly's intent was to show the difference between the NameTypes - the use of "sross" there is accidental leaking of their hostname.  So this should work:

```python
>>> import gssapi, socket
>>> FQDN = socket.getfqdn()
>>> 
>>> server_hostbased_name = gssapi.Name(f"HTTP@{FQDN}", name_type=gssapi.NameType.hostbased_service)
>>> server_hostbased_name
Name(b'HTTP@seton.mivehind.net', <OID 1.2.840.113554.1.2.1.4>)
>>> server_name = gssapi.Name(f"HTTP/{FQDN}@")
>>> server_name
Name(b'HTTP/seton.mivehind.net@', None)
>>> 
>>> server_name == server_hostbased_name
False
>>> server_canon_name = server_name.canonicalize(gssapi.MechType.kerberos)
>>> server_hostbased_canon_name = server_hostbased_name.canonicalize(gssapi.MechType.kerberos)
>>> server_canon_name == server_hostbased_canon_name
True
>>> 
```

So that suggests a better approach, I think.

PR #210, opened 2020-06-04
Support IOV functions on macOS

jborean93:
The DCE IOV functions are not exported in any public header on macOS which means the `ext_dce` functions cannot be used on that platform using the normal way. The actual framework has actually implemented these functions but they need to be loaded dynamically.

This PR uses [dlsym](https://linux.die.net/man/3/dlsym) to lookup the symbol of the GSS.Framework binary and get the address of the function based on the symbol name. The symbols names does not seem to have changed ever since the release of `GSS.Framework` and if they are removed/renamed in a future version the worst thing to happen is for the functions to no longer be available just like before.

One part I'm not sure on is whether we can use the Cython [compile time definitions](https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#compile-time-definitions). If I understand the process correctly we create the sdist with the C files already generated which means the compile time directives won't be in place for macOS. As I see it we could either

* Generate a wheel specifically for macOS and the various Python versions that are supported, or
* Find another way to alternatively define the functions that are done when the C files are compiled and not when Cython generates the C files


simo5:
Uuuggh hairy ....

jborean93:
You're telling me, unfortunately WinRM message encryption requires the IOV functions as it needs the header/trailer length to encode in the exchange messages. Without something stupid like this we can't do that on macOS.

jborean93:
Thanks for replying @frozencemetery. I'll try and have a look at what will go into creating the wheels when I find some time in the future, can't guarantee I'll get to it anytime soon unfortunately.

One question I do have with the sdist that is currently created and published to pypi. Is the cython build just run on a system that has all the extensions present or can Cython create the .c files without the actual C headers potentially being present on the system?

frozencemetery:
Apologies, I didn't mean to make it sound like I'm dumping a lot of work on you.  Creating and uploading wheels should be fairly simple - we have logic for doing it on Windows that should translate, especially https://github.com/pythongssapi/python-gssapi/blob/master/ci/release-win.sh#L11-L13

Inspection suggests that the C files are part of the sdist - the pyx files aren't shipped there.  (Our build logic detects this case.)  It happens that all extensions are present on our build system, but the C files are generated regardless.  When running `pip install` without a wheel, krb5 headers and a C compiler need to be present because the check for whether extensions are present is "does this build and import".

Issue #132, opened 2017-11-28
Add support for GSS_C_DELEG_POLICY_FLAG


frozencemetery:
It's been present in MIT since krb5-1.8, and was added to heimdal at about the same time [the commit message](https://github.com/krb5/krb5/commit/45875a4d7bbd6bb8a943572d84fef5ca2bb18291).

I can't find a standard for it, but the Heimdal wrote up [a piece on its use](http://blog.h5l.org/2008/10/ok-as-delegate-and-gss-api.html).

DirectXMan12:
especially since there's no standard, we'll want to make sure we have good docs in our docs for it.  Otherwise, we'd probably just want it to be its own extension (the tiniest extension).

simo5:
@DirectXMan12 you want to copy in tree the Heimdal write up ?

DirectXMan12:
That's probably fine.  I just like to make it so that our documentation can be used as a centralized place to see the basic mechanics of the APIs we support, since, for many of them, there's no easier place to go (I had to cobble a lot together from the krb5 wiki based on old info, for instance).

PR #218, opened 2020-07-09
Added ok_as_delegate flags

jborean93:
Adds support for `GSS_C_DELEG_POLICY_FLAG` and some docs around it. Currently verifying the behaviour on MIT before finalising the commit message and docs.

Fixes https://github.com/pythongssapi/python-gssapi/issues/132

### DIFF
--- a/docs/source/gssapi.rst
+++ b/docs/source/gssapi.rst
@@ -65,6 +65,34 @@ imported in the high-level API :mod:`gssapi` module:
 .. autoclass:: gssapi.RequirementFlag
     :show-inheritance:
 
+The ``ok_as_delegate`` flag corresponds to the C level flag
+``GSS_C_DELEG_POLICY_FLAG``. This flag is similar to ``delegate_to_peer``
+except it only delegates if the KDC delegation policies for the service
+principal allow it to use delegation. This is typically used on Microsoft
+domain environments to control whether constrained or unconstrained delegation
+is allowed for a service principal. By setting this flag, the delegation
+process follows the same behaviour as delegation on SSPI/Windows.
+
+Here are the four cases when either of these flags are set or not.
+
+Neither flag set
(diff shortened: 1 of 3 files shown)


docs/source/gssapi.rst
```diff
@@ -65,6 +65,34 @@ imported in the high-level API :mod:`gssapi` module:
...
+Here are the four cases when either of these flags are set or not.
+
+Neither flag set
+   No delegation occurs.
+
+delegate_to_peer
+   Always try to delegate regardless of the KDC delegation policies.
+   ``delegate_to_peer`` is set in the return flags if successful.
+
+ok_as_delegate
```
jborean93, line 85: This scenario does not work as documented on MIT. Trying to get some clarification from the krb5 devs to see what they say before updating the docs.

jborean93:
Have tested the behaviour of this flag and have documented the results at https://gist.github.com/jborean93/cdc6cef5e7540ade39912ca08020706b. Waiting on some further info before updating the docs https://github.com/pythongssapi/python-gssapi/pull/218#discussion_r452534896.

docs/source/gssapi.rst
jborean93: Turns out I was doing some faulty testing, after trying again I found that MIT was acting just like Heimdal and what was in the documents. No idea what I did wrong the first time but it's good they act the same :)

jborean93:
@frozencemetery sorry for the false alarm, turns out MIT does act the same as Heimdal I must have done something wrong in my tests. The PR is ready for review.

I ended up calling the `RequirementFlag` `ok_as_delegate` as that matches the name in [RFC4120](https://tools.ietf.org/html/rfc4120#section-2.8) rather than the name of the GSSAPI macro name which is more a software implementation detail.

Issue #221, opened 2020-07-14
six shouldn't be required anymore


jborean93:
### What went wrong?
Now that python 2 is dropped you shouldn't need the `six` package at all.

frozencemetery:
Correct :)

jborean93:
Done https://github.com/pythongssapi/python-gssapi/pull/222.

PR #219, opened 2020-07-10
tests: Drop dependency on python-should_be

stanislavlevin:
Migrate should_be assertions to unittest(stdlib) ones.
This requires Python3.2+.

Fixes: https://github.com/pythongssapi/python-gssapi/issues/215

PR #222, opened 2020-07-14
Removed six package and other older py cleanup values

jborean93:
Six was only required when the code was trying to be py2 compatible. That's no longer the case so we can remove that requirement.

Fixes https://github.com/pythongssapi/python-gssapi/issues/221


README.txt
```diff
@@ -38,7 +38,7 @@ Basic
 
 * Python 3.6+ (older releases support older versions, but are unsupported)
 
-* the `six` and `decorator` python packages
+* the `decorator` python packages
```
frozencemetery, line 41: Package, singular.

tox.ini
```diff
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py27,py33,py34,py35,py36,py37
+envlist = py36,py37,py38
```
frozencemetery, line 7: python-3.8 won't work yet; we need to figure out and merge #219 for that.  This PR will cause a merge conflict with that one, but it should be easily sorted once this merges.  Since we don't use tox ourselves, this is probably fine.

gssapi/_utils.py
```diff
@@ -103,12 +102,12 @@ def set_encoding(enc):
...
     """Encodes any relevant strings in a dict"""
     def enc(x):
-        if isinstance(x, six.text_type):
+        if isinstance(x, str):
             return x.encode(_ENCODING)
         else:
             return x
 
-    return dict((enc(k), enc(v)) for k, v in six.iteritems(d))
+    return dict((enc(k), enc(v)) for k, v in iter(d.items()))
```
frozencemetery, line 110: This can be more clear with a dict comprehension.

PR #223, opened 2020-08-26
Fix shlex.split on windows

dhirschfeld:
I'm trying to compile this package on Windows for [`conda-forge`](https://conda-forge.org/) so need to pass in the `GSSAPI_LINKER_ARGS` and `GSSAPI_COMPILER_ARGS` env vars. Unfortunately they're specified using the Windows backslash convention for paths which requires passing the `posix=False` argument to `shlex.split`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,8 +37,9 @@ def get_output(*args, **kwargs):
 
 
 # get the compile and link args
+posix = os.name != 'nt'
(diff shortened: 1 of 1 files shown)


aiudirog:
I have no problems with this, sounds like a good change.

dhirschfeld:
Thanks!

PR #232, opened 2020-11-19
Fix freebsd

frozencemetery:
- Handle GSS_C_DELEG_POLICY_FLAG not being defined by Heimdal in base
- Support Heimdal from ports preferentially to that from base, if available


frozencemetery:
Added a third commit to fix our feature detection by finding the correct library.  Ready for review.

PR #234, opened 2020-11-30
Mention the env var not the Python var

simmel:
When logging the error mention which environment variable to use to
disable detection. This also matches GSSAPI_MAIN_LIB.


simmel:
It's early and my coffe cup isn't empty yet but it sure took a while for me to find out why detection wasn't disabled ; P

frozencemetery (approved):
Thanks for the patch!

Issue #235, opened 2020-12-09
Unable to build/install on macOS Big Sur 11.0.1


simmel:
### What went wrong?

Not possible to install `pip3 install gssapi` nor build via `./setup.py build` on macOS Big Sur 11.0.1:
```terminal
$ pip3 install ~/src/python-gssapi
Processing /Users/simlu/src/python-gssapi
    ERROR: Command errored out with exit status 1:
     command: /Users/simlu/src/sadkfjaksdlf/bin/python3 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/tmp/pip-req-build-m5_l08gj/setup.py'"'"'; __file__='"'"'/private/tmp/pip-req-build-m5_l08gj/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /private/tmp/pip-pip-egg-info-skp7_31t
         cwd: /private/tmp/pip-req-build-m5_l08gj/
    Complete output (6 lines):
    Cython not found, building from C files...
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/private/tmp/pip-req-build-m5_l08gj/setup.py", line 196, in <module>
        raise Exception("Could not find main GSSAPI shared library.  Please "
    Exception: Could not find main GSSAPI shared library.  Please try setting GSSAPI_MAIN_LIB yourself or setting GSSAPI_SUPPORT_DETECT to 'false'
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
WARNING: You are using pip version 20.2.3; however, version 20.3.1 is available.
You should consider upgrading via the '/Users/simlu/src/sadkfjaksdlf/bin/python3 -m pip install --upgrade pip' command.
```

```terminal
$ python3 ./setup.py build
Cython not found, building from C files...
Traceback (most recent call last):
  File "/Users/simlu/src/python-gssapi/./setup.py", line 196, in <module>
    raise Exception("Could not find main GSSAPI shared library.  Please "
Exception: Could not find main GSSAPI shared library.  Please try setting GSSAPI_MAIN_LIB yourself or setting GSSAPI_SUPPORT_DETECT to 'false'
```

### How do we reproduce?

Build on Big Sur?

[...]

PR #236, opened 2020-12-10
Fix GSSAPI detection on macOS Big Sur

jborean93:
CPython on macOS Big Sur has a bug in the ctypes.util.find_library function.
This has been fixed recently but has not made its way to an actual release
yet. This change adds a fallback GSS framework path to try incase ctypes fails
to find the path.

Fixes https://github.com/pythongssapi/python-gssapi/issues/235

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -162,6 +162,14 @@ if ENABLE_SUPPORT_DETECTION:
     main_path = ""
     if main_lib is None and osx_has_gss_framework:
         main_lib = ctypes.util.find_library('GSS')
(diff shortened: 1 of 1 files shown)


simmel:
Also builds for me. Great work!

frozencemetery (approved):

frozencemetery:
While GitHub Actions lists a label for Big Sur, it doesn't actually run on Big Sur right now, so CI isn't yet possible for this platform.

PR #238, opened 2021-01-12
Document how to use a non-detected Kerberos lib

simmel:
Not sure about the placing or wording.

You can specify it both when you install it via pip but also when you install it from source. What do you think?

I was looking for this docs when I first tried to troubleshoot the macOS 11 issues but I couldn't figure it out. Thanks @jborean93 for mentioning it!

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -68,6 +68,16 @@ Easy Way
 
     $ pip install gssapi
 
+You can also use a specific Kerberos library by defining the `GSSAPI_*`
+environment variables:
+
+.. code-block:: bash
+
+    $ export PATH=/path/to/where/krb5-config/is:$PATH
+    $ export GSSAPI_LINKER_ARGS="$(krb5-config --libs gssapi)"
+    $ export GSSAPI_COMPILER_ARGS="$(krb5-config --cflags gssapi)"
+    $ export GSSAPI_MAIN_LIB="/path/to/where/libgss{{api,-3}.dll,api.{so,dylib}}/is"
+
 From the Git Repo
 -----------------
 


README.txt
```diff
@@ -68,6 +68,16 @@ Easy Way
...
 
+You can also use a specific Kerberos library by defining the `GSSAPI_*`
+environment variables:
+
+.. code-block:: bash
+
+    $ export GSSAPI_LINKER_ARGS="$(krb5-config --libs gssapi)"
+    $ export GSSAPI_COMPILER_ARGS="$(krb5-config --cflags gssapi)"
+    $ # Too many variants, search for GSSAPI_MAIN_LIB in setup.py
+    $ export GSSAPI_MAIN_LIB="$PATH_TO/lib/yourkerberoslib.ext"
```
jborean93, line 79: This should be `PATH_TO/lib/libgssapi.so` with a note saying the extension is `.dylib` on macOS. You might also want to mention that `krb5-config` selected in `PATH` should be for the Kerberos library you wish to link against.

simmel: There are other usecases on Windows; mingw, MIT Kerberos for Windows etc. Hence the comment. It became too messy to list them all.

Can the comment be improved somehow and still be brief?

Re: PATH, yes. Or maybe specify hard path e.g. `/path/to/you/krb5-config`. What's best?

jborean93: > There are other usecases on Windows; mingw, MIT Kerberos for Windows etc. Hence the comment. It became too messy to list them all.

I can't say I use this with Windows so I can't really speak more, I just think it should be more obvious that the default is `libgssapi.so` for Linux and `libgssapi.dylib` for macOS as per https://github.com/pythongssapi/python-gssapi/blob/a7de108528053417459e6b7ef64aad1b89ab8415/setup.py#L186-L192.

>  Or maybe specify hard path e.g. /path/to/you/krb5-config

You need to make sure it's first in `PATH`. Your linker and compiler args can use the full path but [setup.py calls](https://github.com/pythongssapi/python-gssapi/blob/a7de108528053417459e6b7ef64aad1b89ab8415/setup.py#L129) `krb5-config gssapi --prefix` so you need to ensure the correct `krb5-config` is selected there.

simmel: > You need to make sure it's first in PATH. Your linker and compiler args can use the full path but setup.py calls krb5-config gssapi --prefix so you need to ensure the correct krb5-config is selected there.

I totally missed that! Fixing.

README.txt
```diff
@@ -68,6 +68,16 @@ Easy Way
...
 
+You can also use a specific Kerberos library by defining the `GSSAPI_*`
+environment variables:
+
+.. code-block:: bash
+
+    $ export PATH=/path/to/where/krb5-config/is:$PATH
+    $ export GSSAPI_LINKER_ARGS="$(krb5-config --libs gssapi)"
+    $ export GSSAPI_COMPILER_ARGS="$(krb5-config --cflags gssapi)"
+    $ export GSSAPI_MAIN_LIB="/path/to/where/libgss{{api,-3}.dll,api.{so,dylib}}/is"
```
simmel, line 79: > I can't say I use this with Windows so I can't really speak more, I just think it should be more obvious that the default is libgssapi.so for Linux and libgssapi.dylib for macOS as per 

I really don't think this is clearer @jborean93

frozencemetery:
In most cases, the issue is simply locating krb5-config, which will produce the right compiler/linker flags for us.  Setting the compile/link/library flags manually shouldn't be required (outside of porting work).  If that's not the behavior, we should fix it :)

I think we should reflect that in any install instructions: that mostly one just sets PATH, and mention that the other env vars exist.  (It might be clearest not to even give usage examples for them, since it's potentially confusing.)

simmel:
On Wed, 2021-01-13 at 08:13:32 -0800, Robbie Harwood wrote:
>In most cases, the issue is simply locating krb5-config, which will produce the right compiler/linker flags for us.  Setting the compile/link/library flags manually shouldn't be required (outside of porting work).  If that's not the behavior, we should fix it :)

That's a great attitude! But I don't think that's possible everywhere, 
especially for GSSAPI_MAIN_LIB.

Look at 
https://github.com/pythongssapi/python-gssapi/blob/a7de10/setup.py#L158-L208 
no where is it using PATH or krb5-config to locate the main lib (and I 
don't think it could without guessing a lot). So that means:
$ export PATH=/usr/local/opt/heimdal/bin:$PATH
$ python3 setup.py build
[…]
$ otool -L build/lib.macosx-11-x86_64-3.9/gssapi/raw/mech_krb5.cpython-39-darwin.so
build/lib.macosx-11-x86_64-3.9/gssapi/raw/mech_krb5.cpython-39-darwin.so:
        /System/Library/Frameworks/GSS.framework/Versions/A/GSS (compatibility version 1.0.0, current version 1.0.0)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.60.1)

will never work.

For GSSAPI_COMPILER_ARGS we see at 
https://github.com/pythongssapi/python-gssapi/blob/a7de10/setup.py#L111-L122
that for macOS and Windows (winkrb_path is set 
https://github.com/pythongssapi/python-gssapi/blob/a7de10/setup.py#L52-L65 
) we have statically specified where to look. Well, Windows does fall 
back to looking at PATH where kinit is and use that path but I'm not 
sure that counts as being able to specify it dynamically (since you 
might have MIT Kerberos installed but what to try a new version located 
at another place).

For GSSAPI_LINKER_ARGS it's basically the same as for 
GSSAPI_COMPILER_ARGS:
https://github.com/pythongssapi/python-gssapi/blob/a7de10/setup.py#L95-L109

Maybe removing the "else" in those if-statements and always running 
krb5-config will work but I'm not so sure of that either.

README.txt
jborean93: My main comment was to do document `libgssapi.so` and just add a comment saying this could be `.dylib` or something else on other platforms.

simmel:
stale bot is nice (for people like me who are forgetful ; )

simmel:
@jborean93 is my force push available in this PR or do I need to open a new one since stalebot closed it?

While helping a co-worker getting python-gssapi to build on his Mac we tried to only use `GSSAPI_COMPILER_ARGS` and `GSSAPI_LINKER_ARGS` and that works.

jborean93:
I’m not sure sorry, that’s a question for @frozencemetery

frozencemetery:
> is my force push available in this PR or do I need to open a new one since stalebot closed it?

I can't reopen it (the button is greyed out) - it looks like you might've force-pushed already?  If you're not able to reopen either, I think a new PR would be in order.

Issue #241, opened 2021-04-05
variable referenced before assignment. 


testtubeme:
https://github.com/pythongssapi/python-gssapi/blob/a7de108528053417459e6b7ef64aad1b89ab8415/gssapi/raw/misc.pyx#L312-L313

needs 
```
except ValueError as e:
```

frozencemetery:
Hi, if you'd like this fixed sooner, feel free to submit a pull request :)  Otherwise I'll get to it when I have a bit more time to look.

PR #245, opened 2021-04-30
Fix variable referenced before assignment when decoding status_code msg

jborean93:
Fixes https://github.com/pythongssapi/python-gssapi/issues/241


frozencemetery (approved):
Thanks!

PR #248, opened 2021-05-25
Configure Windows at import time to use the 32-bit kerbeors on a 32-bit python

asheshv:
This change will allow python-gssapi on a 32-bit python application.

### DIFF
--- a/gssapi/_win_config.py
+++ b/gssapi/_win_config.py
@@ -8,6 +8,7 @@ down the import tree.
 
 import os
 import shutil
+import sys
 import ctypes
 
 #: Path to normal KfW installed bin folder
@@ -22,7 +23,10 @@ KFW_DL = "https://web.mit.edu/KERBEROS/dist"
 def kfw_available():
     """Return if the main GSSAPI DLL for KfW can be loaded"""
     try:  # to load the main GSSAPI DLL
-        ctypes.WinDLL('gssapi64.dll')
+        if sys.maxsize > 2**32:
+            ctypes.WinDLL('gssapi64.dll')
+        else:
+            ctypes.WinDLL('gssapi32.dll')
     except OSError:  # DLL is not in PATH
         return False
     else:  # DLL is in PATH, everything should work
--- a/setup.py
+++ b/setup.py
@@ -39,6 +39,11 @@ def get_output(*args, **kwargs):
 # get the compile and link args
 kc = "krb5-config"
 posix = os.name != 'nt'
+
+# Per https://docs.python.org/3/library/platform.html#platform.architecture
+# this is the preferred way of determining "64-bitness".
+is64bit = sys.maxsize > 2**32
+
 link_args, compile_args = [
     shlex.split(os.environ[e], posix=posix) if e in os.environ else None
     for e in ['GSSAPI_LINKER_ARGS', 'GSSAPI_COMPILER_ARGS']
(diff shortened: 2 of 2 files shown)


gssapi/_win_config.py
```diff
@@ -17,12 +18,16 @@
 )
 #: Download location for KfW
 KFW_DL = "https://web.mit.edu/KERBEROS/dist"
+is64bits = sys.maxsize > 2**32
```
frozencemetery, line 21: This is only referenced once, so it's not really clearer to have it be its own variable.

setup.py
```diff
@@ -39,6 +39,14 @@ def get_output(*args, **kwargs):
...
 kc = "krb5-config"
 posix = os.name != 'nt'
+
+# According to the Python documentation, in order to get the '64-bitness' of
+# the current interpreter, it is more reliable to query the sys.maxsize
+# attribute:
+#
+# Reference:
+# https://docs.python.org/3/library/platform.html#platform.architecture
+is64bits = sys.maxsize > 2**32
```
frozencemetery, line 49: (Same here.)

frozencemetery (changes requested):
Hi, thanks for your interest in this project.  This seems useful, so I do intend to merge support.  Couple things first:

- You'll need to pass CI (doesn't look like anything major is broken).
- Your commit message subject is too long.  I can fix this before merging, but if you'd like to yourself, I suggest reading [this post](https://chris.beams.io/posts/git-commit/) for an idea.

setup.py
asheshv: On closer look at the code, we're using `sys.maxsize > 2**32` for testing 64-bit code execution in the setup script.
Hence - using this variable at all those places now.

gssapi/_win_config.py
asheshv: I do agree about using a variable instead of the logic in general.

But - using the statement `sys.maxsize > 2**32` will puzzle people in long term, what's going on here.
And - that's the reason - I have used a variable here.

 I update the pull request to remove the variable use here.

asheshv:
> Hi, thanks for your interest in this project. This seems useful, so I do intend to merge support. Couple things first:
> 
> * You'll need to pass CI (doesn't look like anything major is broken).
I've checked last CI, something was broken in the last commit.
It is not fixed as well.

> * Your commit message subject is too long.  I can fix this before merging, but if you'd like to yourself, I suggest reading [this post](https://chris.beams.io/posts/git-commit/) for an idea.
That would be helpful.

frozencemetery:
@aiudirog Since we don't have a way to test 32-bit Windows in CI right now, would you mind double-checking this?

aiudirog:
This looks good to me, but I think we should add CI support for 32bit, at least on Windows. I think it actually should be fairly easy with a slight modification to the matrix. Give me few moments to try something.

aiudirog:
I can't push  directly to this branch, so we'll go old school with a patch file.

[0001-CI-Build-and-release-on-32bit-Windows-Python.txt](https://github.com/pythongssapi/python-gssapi/files/6541330/0001-CI-Build-and-release-on-32bit-Windows-Python.txt)

-----

Also, GitHub, why do you not support .patch/.diff files....?

frozencemetery:
Thanks, applied it.  I think 8 is probably overkill for Windows - let me see if I can pare that down...

aiudirog:
Looks like the i386 folder isn't available when kerberos is installed through chocolatey. Maybe we need to find a 32bit specific one?

frozencemetery:
Beyond me, sorry.  I think adding 32-bit CI isn't a bad idea if it can be made to work.  I think you can open a separate PR using the commit here as basis?

aiudirog:
That should work, I'll see what I can do.

PR #251, opened 2021-05-31
Add support for GSS_C_CHANNEL_BOUND_FLAG

steffen-kiess:
MIT krb5 version 1.19 will set GSS_C_CHANNEL_BOUND_FLAG when channel
binding succeeds. This will cause gssapi to fail with the error message:

ValueError: 2048 is not a valid RequirementFlag

Add support for GSS_C_CHANNEL_BOUND_FLAG to fix this.


frozencemetery (approved):
Thanks!

PR #253, opened 2021-06-02
Enable specification of krb5-config via environment variable

### LINKED ISSUE #250: Enable krb5-config auto-detect override in setup.py on FreeBSD
a667468 introduced new detection logic to prefer more modern Kerberos libraries from the ports tree over the older Heimdal in base. This is an excellent choice for those who wish to install directly. As the maintainer of the security/py-gssapi port, I need to be able to A) build deterministically, and B) respect the OPTIONs set by a port builder in a way that is insensitive to the state of the build environment. That is, I need to be able to override any auto-detection.

In the current version of the port ([1.6.9](https://cgit.freebsd.org/ports/tree/security/py-gssapi?id=f70f34331a4a388b8f720a44713466221b64cdae)), I accomplish this by patching `setup.py` with the absolute path of whatever `krb5-config` is applicable to the GSSAPI library selected by the port builder (`$KRB5CONFIG` per the ports machinery).

I will be glad to draft a patch. First, I wanted to get feedback from the project to see if this would be welcome. If not, I can adapt the port-build-time patch to the new code. If so, is there an approach I should prefer/avoid for the purpose of general utility and maintainability? An environment variable, perhaps?

neirbowj:
Closes #250

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,11 +38,18 @@ def get_output(*args, **kwargs):
 
 # get the compile and link args
 kc = "krb5-config"
+autodetect_kc = True
 posix = os.name != 'nt'
 
 # Per https://docs.python.org/3/library/platform.html#platform.architecture
 # this is the preferred way of determining "64-bitness".
 is64bit = sys.maxsize > 2**32
+
+kc_env = 'GSSAPI_KRB5CONFIG'
+if kc_env in os.environ:
+    kc = os.environ[kc_env]
+    autodetect_kc = False
+    print(f"Using {kc} from env")
 
 link_args, compile_args = [
     shlex.split(os.environ[e], posix=posix) if e in os.environ else None
@@ -77,7 +84,7 @@ if os.name == 'nt':
     except ValueError:
         cygwinccompiler.get_msvcr = lambda *a, **kw: []
 
-if sys.platform.startswith("freebsd"):
+if sys.platform.startswith("freebsd") and autodetect_kc:
     # FreeBSD does $PATH backward, for our purposes.  That is, the package
     # manager's version of the software is in /usr/local, which is in PATH
     # *after* the version in /usr.  We prefer the package manager's version


neirbowj:
Build look good with py36, py37, py38 (default), and py39 on FreeBSD 12.2-RELEASE amd64 against:

- [Heimdal in base](https://pkg.saltant.net/poudriere/build.html?mastername=122amd64-default-gssbase&build=2021-06-01_20h12m55s)
- [Heimdal in ports](https://pkg.saltant.net/poudriere/build.html?mastername=122amd64-default-gssheimdal&build=2021-06-01_20h14m29s)
- [MIT in ports](https://pkg.saltant.net/poudriere/build.html?mastername=122amd64-default-gssmit&build=2021-06-01_20h15m47s)

setup.py
```diff
@@ -38,12 +38,19 @@ def get_output(*args, **kwargs):
...
 # get the compile and link args
 kc = "krb5-config"
+autodetect_kc = True
 posix = os.name != 'nt'
 
 # Per https://docs.python.org/3/library/platform.html#platform.architecture
 # this is the preferred way of determining "64-bitness".
 is64bit = sys.maxsize > 2**32
 
+kc_env = 'KRB5CONFIG'
```
jborean93, line 48: This is suspiciously close to `KRB5_CONFIG` which is an env var that MIT (maybe Heimdal as well) uses to specify the `krb5.conf` file to use. Would it be better to use the `GSSAPI_` prefix we use for the other env var configs like `GSSAPI_KRB5CONFIG` to avoid any confusion?

neirbowj: That's good feedback. I chose `KRB5CONFIG` because [that's what the FreeBSD ports machinery uses for the purpose](https://cgit.freebsd.org/ports/tree/Mk/Uses/gssapi.mk?id=7229781c12bb3ffe8a8f9fa0f6fb2805802951e2#n138). Since my intent with this patch is to contribute something more generally useful, I would be glad to tailor to suit.

frozencemetery:
Hmm, okay.  I think I had a perhaps more invasive change in mind.

Right now we've got `$GSSAPI_LINKER_ARGS` (overrides `link_args`, but can be derived from `krb5_config`), `$GSSAPI_COMPILER_ARGS` (overrides `compile_args`, but can be derived from `krb5_config`), and `$GSSAPI_MAIN_LIB` (overrides `main_lib`, but can be derived from `link_args` and platform goo).  So what I had in mind was:

1. The freebsd stanza should respect `link_args` and `compile_args` if they're already set - this allows for proper overriding with the existing env vars.
2. Optionally, some of the code in setup.py could move around to make more sense, or move into helper functions for clarity.  I think classes is probably overkill, though.
3. As an optional add-on, we add a variable for the location of `krb5_config`.  This is then used to derive `link_args` and `compile_args` if they're not set (and the it makes sense on the platform etc.).

Since I'm being fussy here, please let me know if you'd rather I make those changes :)

neirbowj:
> 1. The freebsd stanza should respect `link_args` and `compile_args` if they're already set - this allows for proper overriding with the existing env vars.

Unless I'm missing something, it already works this way. More to the point, the FBSD logic only affects `kc` (whether to rely on the PATH by default, or call by absolute path), which is the later used to learn `link_args` and `compile_args` if they are not already supplied in the environment.

> 2. Optionally, some of the code in setup.py could move around to make more sense, or move into helper functions for clarity. I think classes is probably overkill, though.

Makes complete sense to me. That's a little more than I think I can sign up to at the moment, though.

> 3. As an optional add-on, we add a variable for the location of krb5_config. This is then used to derive `link_args` and `compile_args` if they're not set (and the it makes sense on the platform etc.).

Now I really feel like I'm missing something. `kc` is a variable with the location of `krb5-config`. It is already used to set `link_args` and `compile_args` if they are not already set, _and_ they are not known in a platform-dependent manner.

frozencemetery:
Hmm, you're right - I managed to confuse myself somewhere in there.  We merely *set* krb5-config, but don't actually invoke it.  So this should be fine then.

frozencemetery:
@jborean93 Before I merge, are you okay with the new variable name?

jborean93:
My apologies @frozencemetery just saw this, the new name is fine with me.

PR #254, opened 2021-06-22
Document how to use a non-detected Kerberos lib

simmel:
Re-PR of #238

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -68,6 +68,12 @@ Easy Way
 
     $ pip install gssapi
 
+You can also use a specific Kerberos library by adding the path where it's `krb5-config` binary is:
+
+.. code-block:: bash
+
+    $ export PATH=/path/to/where/krb5-config/is:$PATH
+
 From the Git Repo
 -----------------
 


simmel:
@frozencemetery

frozencemetery:
> @frozencemetery

No need to worry, I'm watching the repo and see everything.

I think this is almost right.  To keep it as a pure documentation PR, one possibility is to split it into two sections: first, show how to select krb5-config (PATH is fine; you could optionally mention GSSAPI_KRB5CONFIG but I don't think it's required), and separately show setting GSSAPI_{COMPILER,LINKER}_ARGS.  (Something like "if you also need to override flags detection, you can define ..." maybe.)

My thought here is that this avoids more boilerplate - setup.py will call krb5-config itself in many cases, just not all.  This points at another possibility.  In #238, you mentioned macos - right now that's one of the platforms that won't fall back to calling krb5-config.  But we could also just change that - see 
https://github.com/pythongssapi/python-gssapi/commit/4c23c752d664be2ba63f086fb891e12176b5fe80 which changed that behavior for freebsd.

simmel:
> > @frozencemetery
> 
> No need to worry, I'm watching the repo and see everything.

All seeing eye, huh? = )
 
> I think this is almost right. To keep it as a pure documentation PR, one possibility is to split it into two sections: first, show how to select krb5-config (PATH is fine; you could optionally mention GSSAPI_KRB5CONFIG but I don't think it's required), 

I can't find any references to `GSSAPI_KRB5CONFIG`? I don't know what it is.

>and separately show setting GSSAPI_{COMPILER,LINKER}_ARGS. (Something like "if you also need to override flags detection, you can define ..." maybe.)

I'm pretty sure if people needs to change them they know how they work. Otherwise new users using macOS just gets into the thought train of "I can define this, but do I have to? How do I know? I'm gonna open an issue and ask".
 
> My thought here is that this avoids more boilerplate - setup.py will call krb5-config itself in many cases, just not all. This points at another possibility. In #238, you mentioned macos - right now that's one of the platforms that won't fall back to calling krb5-config. But we could also just change that - see
> [4c23c75](https://github.com/pythongssapi/python-gssapi/commit/4c23c752d664be2ba63f086fb891e12176b5fe80) which changed that behavior for freebsd.

That would be nice but is outside of the scope of this PR.

frozencemetery:
> I can't find any references to GSSAPI_KRB5CONFIG? I don't know what it is.

Pull/rebase; it's recently added to main.

> I'm pretty sure if people needs to change them they know how they work. Otherwise new users using macOS just gets into the thought train of "I can define this, but do I have to? How do I know? I'm gonna open an issue and ask".

Well... that's probably true *now*, but it's not going to be true once your PR which adds mention of the merges.

simmel:
Like that?

PR #258, opened 2021-08-03
Support DCE IOV functions on macOS

jborean93:
The DCE IOV functions on macOS are not exported by any public header on
the GSS.Framework. This PR defines some macros in the C file that when
compiled against the GSS.Framework will alias the private symbols to the
ones expected by GSSAPI. While the symbols are considered to be private
they haven't changed across any macOS version since the introduction of
GSS.Framework and the inclusion of ext_dce is still dependent on whether
the symbol is present at compile time.

This allows users of this library on macOS to better interop with
Windows SSPI message encryption which typically require the IOV wrapping
functions that were previously unavailable.

This is round 2 from https://github.com/pythongssapi/python-gssapi/pull/210 but I'm a lot more happier with this approach as the macros are defined in the `.c` file for the sdist allowing this to work on macOS without requiring a wheel. It is also not relying on `dlopen/dlsym` at runtime to get the path to the GSS.Framework library which is a big plus.


gssapi/raw/ext_dce.pyx
```diff
@@ -8,15 +8,34 @@ from gssapi.raw.sec_contexts cimport SecurityContext
...
+    from gssapi.raw.ext_dce_aead import wrap_aead, unwrap_aead
+except ImportError:
+    pass
+
 
 cdef extern from "python_gssapi_ext.h":
+    """
+    #ifdef OSX_HAS_GSS_FRAMEWORK
+    #define gss_wrap_iov(t, u, v, w, x, y, z) \
+    __ApplePrivate_gss_wrap_iov(t, u, v, w, x, y, z)
```
frozencemetery, line 29: I think it would be more clear to indent the continuation lines, unless that breaks something.

gssapi/raw/python_gssapi_ext.h
```diff
@@ -1,5 +1,43 @@
 #ifdef OSX_HAS_GSS_FRAMEWORK
 #include <GSS/GSS.h>
+
+// GSS includes these in gssapi_private.h but these aren't present on the host
```
frozencemetery, line 4: So far we've been loosely conforming to [krb5 style](https://k5wiki.kerberos.org/wiki/Coding_style/Formatting).  Because I find it easier to explain that visually, this first function would become:

```C
/* GSS includes these in gssapi_private.h but these aren't present on the
 * host, so just explicitly define them. */
GSSAPI_LIB_FUNCTION OM_uint32 GSSAPI_LIB_CALL
__ApplePrivate_gss_unwrap_iov(__nonnull OM_uint32 *minor_state,
                              __nonnull gss_ctx_id_t context_handle,
                              __nullable int *conf_state,
                              __nullable gss_qop_t *qop_state,
                              __nonnull gss_iov_buffer_desc *iov,
                              int iov_count);
```

(note that's all spaces, not tabs).

gssapi/raw/python_gssapi_ext.h
```diff
@@ -1,5 +1,43 @@
...
+// so just explicitly define them.
+GSSAPI_LIB_FUNCTION OM_uint32 GSSAPI_LIB_CALL
+__ApplePrivate_gss_unwrap_iov (
+	OM_uint32 * __nonnull minor_status,
```
frozencemetery, line 8: The nullbility checks appear to have been added to XCode in 6.3, which if I understand the decoder ring correctly, corresponds to 10.9.  At almost five years unsupported, that should be fine as a minimum version :)

However, neither clang nor gcc have support for these (though they appear planned for the upcoming clang-13).  Would this limit our ability to build in certain MacOS setups?  (And, if so, do we care?)

setup.py
```diff
@@ -122,7 +122,7 @@ def get_output(*args, **kwargs):
 
 if compile_args is None:
     if osx_has_gss_framework:
-        compile_args = ['-framework', 'GSS', '-DOSX_HAS_GSS_FRAMEWORK']
+        compile_args = ['-DOSX_HAS_GSS_FRAMEWORK']
```
frozencemetery, line 125: Why don't we need `-framework GSS` anymore?

jborean93:
Oh and the change https://github.com/pythongssapi/python-gssapi/pull/258/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R125 is unrelated but clang will emit warning saying this is unused as compiler args. You only need this during the linking phase.

gssapi/raw/ext_dce.pyx
```diff
@@ -8,15 +8,34 @@ from gssapi.raw.sec_contexts cimport SecurityContext
...
+# Kept for backwards compatibility - functions used to be declared here
+try:
+    from gssapi.raw.ext_dce_aead import wrap_aead, unwrap_aead
+except ImportError:
+    pass
+
 
 cdef extern from "python_gssapi_ext.h":
+    """
+    #ifdef OSX_HAS_GSS_FRAMEWORK
```
frozencemetery, line 27: You can just shove C in here?  Amazing

frozencemetery (commented):
In general I don't see any problems with this approach.  Some commentary inline.

gssapi/raw/ext_dce.pyx
frozencemetery: Also: if this is actually just C, there shouldn't be a need to specify the arguments at all - one can just do e.g. `#define gss_wrap_iov __ApplePrivate_gss_wrap_iov`

frozencemetery:
> clang will emit warning saying this -framework ... is unused as compiler args. You only need this during the linking phase.

Makes sense.  Should be mentioned in the commit description (unless you think it warrants its own commit).

gssapi/raw/python_gssapi_ext.h
jborean93: I don't mind whatever way, this is literally a copy of the headers from https://opensource.apple.com/source/Heimdal/Heimdal-597.121.1/lib/gssapi/gssapi/gssapi_private.h.auto.html but with the names renamed. The renaming is part of the build process for Apple, they just don't include this header :(

gssapi/raw/python_gssapi_ext.h
jborean93: That's a good point, I'm not really too well versed in how headers actually work but it sounds like we should just have this as the normal krb5 header to just match out declaration in ext_dce.pyx. Do you see any problems with that?

gssapi/raw/ext_dce.pyx
jborean93: Yea, I stumbled across this yesterday and thought it would make this simpler https://cython.readthedocs.io/en/latest/src/userguide/external_C_code.html#including-verbatim-c-code.

jborean93:
>  Should be mentioned in the commit description

I'll amend the commit to include this.

gssapi/raw/ext_dce.pyx
jborean93: Nice, just tested it and it works just fine. This is my C noobness shining through.

gssapi/raw/python_gssapi_ext.h
frozencemetery: So far we've been converting to match project style in the cython, so I would prefer we keep doing that.  If you want to mention more detail about where these declarations come from in the comment here I think that would be fine - up to you.

gssapi/raw/python_gssapi_ext.h
frozencemetery: Are you suggesting dropping the extra qualifiers?  If so, I think that'd work - it's the most cautious way to go.  Even without knowing what XCode's compiler does with them, the most we'd lose is some NULL checking - not great, but what can/can't be NULL seems pretty clear for these functions.

But I also don't know the macOS ecosystem well enough to know if building with a non-XCode compiler is a thing people do anymore :)

jborean93: Yea, considering the cython declarations don't have them I'm sure it will work. I'll give it a shot and see how it goes.

gssapi/raw/python_gssapi_ext.h
jborean93: Sounds good, will do

jborean93:
Thanks @frozencemetery for the review, I've incorporated the changes and have manually tested again and things work just fine.

Issue #143, opened 2018-01-16
Restore lost docs on result tuples


DirectXMan12:
We seem to have lost docs on the result tuples (see, for example, 
https://pythongssapi.github.io/python-gssapi/latest/gssapi.raw.html?highlight=addcredresult#gssapi.raw.ext_password_add.add_cred_with_password)

We should really make sure those are back in place so that people can look at what's actually returned.

PR #260, opened 2021-08-05
Tidy up docs and turn warnings into errors

jborean93:
The docs had a few minor issues with return value indentation, missing extension docs, and some other minor problems. There were also over 200 warnings being output during the generation mostly around unambiguous type selection (`gssapi.OID` vs `gssapi.raw.OID`). This fixes those warnings and also enables the Sphinx option to treat any future warnings as errors to stop it from building up again.

Fixes https://github.com/pythongssapi/python-gssapi/issues/143

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -143,9 +143,13 @@ Extensions
 
 In addition to RFC 2743/2744, Python-GSSAPI also has support for:
 
+* RFC 4178 (GSS-API Negotiation Mechanism)
+
 * RFC 5587 (Extended GSS Mechanism Inquiry APIs)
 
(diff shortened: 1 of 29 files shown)


frozencemetery (changes requested):
This seems fine so far, but I think I broke the preferred merge order between your PRs.  Can you rebase this onto main?

jborean93:
@frozencemetery I was sure merging them in any order would have caused a rebase :) Have pushed it to the branch and should be good to go now.

frozencemetery (approved):

Issue #233, opened 2020-11-19
Support FreeBSD in tests?


frozencemetery:
### What went wrong?

It's expected that heimdal isn't supported by tests.  But krb5 should be; however, it fails with what appears to be a k5test problem:

```
EE
======================================================================
ERROR: Failure: TypeError (expected str, bytes or os.PathLike object, not NoneType)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/nose/failure.py", line 39, in runTest
    raise self.exc_val.with_traceback(self.tb)
  File "/usr/local/lib/python3.7/site-packages/nose/loader.py", line 417, in loadTestsFromName
    addr.filename, addr.module)
  File "/usr/local/lib/python3.7/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/usr/local/lib/python3.7/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/home/vagrant/.venv/lib/python3.7/imp.py", line 234, in load_module
    return load_source(name, filename, file)
  File "/home/vagrant/.venv/lib/python3.7/imp.py", line 171, in load_source
    module = _load(spec)
  File "<frozen importlib._bootstrap>", line 696, in _load
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/usr/home/vagrant/python-gssapi/gssapi/tests/test_high_level.py", line 16, in <module>
    import k5test.unit as ktu
  File "/home/vagrant/.venv/lib/python3.7/site-packages/k5test/__init__.py", line 1, in <module>
    from k5test.realm import K5Realm  # noqa
  File "/home/vagrant/.venv/lib/python3.7/site-packages/k5test/realm.py", line 92, in <module>
    'kdb'),
  File "/home/vagrant/.venv/lib/python3.7/posixpath.py", line 80, in join
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not NoneType

======================================================================
ERROR: Failure: TypeError (expected str, bytes or os.PathLike object, not NoneType)
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #262, opened 2021-08-11
Support for running tests against Heimdal in CI

jborean93:
~This requires https://github.com/pythongssapi/k5test/pull/13 to be reviewed/merged/released first as currently it is installing straight from source to verify the tests do in fact work. Once that is done the lines https://github.com/pythongssapi/python-gssapi/pull/262/files#diff-0466ec3bdb3913aba71d897f97c298a0de3f922e14fe75af374e16e8478104c3R111-R114 can be removed from this PR.~

I initially looked at running these tests on Windows as well but it doesn't look like MIT Kerberos on Windows provides a kdc service so I dropped that idea.

Should fix https://github.com/pythongssapi/python-gssapi/issues/233 - at least combined with the k5test PR.

### DIFF
--- a/ci/run-on-linux.sh
+++ b/ci/run-on-linux.sh
@@ -3,7 +3,7 @@
 # If we try to use a normal Github Actions container with
 # github-pages-deploy-action, it will fail due to inability to find git.
 
-docker run -h test.box \
+docker run -h test.krbtest.com \
        -v `pwd`:/tmp/build -w /tmp/build \
        -e KRB5_VER=${KRB5_VER:-mit} \
        -e FLAKE=${FLAKE:no} \
--- a/gssapi/tests/test_high_level.py
+++ b/gssapi/tests/test_high_level.py
@@ -18,7 +18,9 @@ import k5test as kt
 
 
 TARGET_SERVICE_NAME = b'host'
-FQDN = socket.getfqdn().encode('utf-8')
+FQDN = (
+    'localhost' if sys.platform == 'darwin' else socket.getfqdn()
+).encode('utf-8')
 SERVICE_PRINCIPAL = TARGET_SERVICE_NAME + b'/' + FQDN
 
 # disable error deferring to catch errors immediately
@@ -124,7 +126,8 @@ class CredsTestCase(_GSSAPIKerberosTestCase):
                  usage='both')
     def test_acquire_by_init(self, str_name, kwargs):
         creds = gsscreds.Credentials(name=self.name, **kwargs)
-        self.assertIsInstance(creds.lifetime, int)
+        if sys.platform != 'darwin':
+            self.assertIsInstance(creds.lifetime, int)
         del creds
 
     @exist_perms(lifetime=30, mechs=[gb.MechType.kerberos],
@@ -137,7 +140,8 @@ class CredsTestCase(_GSSAPIKerberosTestCase):
         creds, actual_mechs, ttl = cred_resp
         self.assertIsInstance(creds, gsscreds.Credentials)
         self.assertIn(gb.MechType.kerberos, actual_mechs)
-        self.assertIsInstance(ttl, int)
+        if sys.platform != 'darwin':
+            self.assertIsInstance(ttl, int)
 
         del creds
 
@@ -165,9 +169,12 @@ class CredsTestCase(_GSSAPIKerberosTestCase):
         self.assertIsNotNone(deleg_creds)
 
         store_res = deleg_creds.store(usage='initiate', set_default=True,
+                                      mech=gb.MechType.kerberos,
                                       overwrite=True)
-        self.assertEqual(store_res.usage, "initiate")
-        self.assertIn(gb.MechType.kerberos, store_res.mechs)
+        # While Heimdal doesn't fail it doesn't set the return values as exp.
+        if self.realm.provider.lower() != 'heimdal':
+            self.assertEqual(store_res.usage, "initiate")
+            self.assertIn(gb.MechType.kerberos, store_res.mechs)
 
         reacquired_creds = gsscreds.Credentials(name=deleg_creds.name,
                                                 usage='initiate')
@@ -187,10 +194,18 @@ class CredsTestCase(_GSSAPIKerberosTestCase):
         initial_creds = gsscreds.Credentials(name=None,
                                              usage='initiate')
 
-        store_res = initial_creds.store(store, overwrite=True)
+        acquire_kwargs = {}
+        expected_usage = 'initiate'
+        if self.realm.provider.lower() == 'heimdal':
+            acquire_kwargs['usage'] = 'initiate'
+            acquire_kwargs['mech'] = gb.MechType.kerberos
+            expected_usage = 'both'
+
+        store_res = initial_creds.store(store, overwrite=True,
+                                        **acquire_kwargs)
         self.assertIsNotNone(store_res.mechs)
         self.assertGreater(len(store_res.mechs), 0)
-        self.assertEqual(store_res.usage, "initiate")
+        self.assertEqual(store_res.usage, expected_usage)
 
         name = gssnames.Name(princ_name)
         retrieved_creds = gsscreds.Credentials(name=name, store=store)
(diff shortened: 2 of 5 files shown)


ci/run-on-linux.sh
```diff
@@ -3,7 +3,7 @@
 # If we try to use a normal Github Actions container with
 # github-pages-deploy-action, it will fail due to inability to find git.
 
-docker run -h test.box \
+docker run -h test.krbtest.com \
```
frozencemetery, line 6: I imagine this broke something?

gssapi/tests/test_high_level.py
```diff
@@ -200,6 +215,7 @@ def test_create_from_other(self):
         raw_creds = gb.acquire_cred(None, usage='accept').creds
 
         high_level_creds = gsscreds.Credentials(raw_creds)
+
```
frozencemetery, line 218: whitespace

gssapi/tests/test_high_level.py
```diff
@@ -381,8 +417,19 @@ def test_sasl_properties(self):
             if mech.description:
                 self.assertIsInstance(mech.description, str)
 
-            cmp_mech = gssmechs.Mechanism.from_sasl_name(mech.sasl_name)
-            self.assertEqual(str(cmp_mech), str(mech))
+            # Heimdal fails with Unknown mech-code on sanon
+            if not (
```
frozencemetery, line 421: Please don't write if statements this way.  Project style is closer to:

```python
            if not (self.realm.provder.lower() == "heimdal" and
                    s = "1.3.6.1.4.1.5322.26.1.110"):
```

gssapi/tests/test_high_level.py
```diff
@@ -381,8 +417,19 @@ def test_sasl_properties(self):
...
+                self.realm.provider.lower() == 'heimdal' and
+                s == '1.3.6.1.4.1.5322.26.1.110'
+            ):
+                cmp_mech = gssmechs.Mechanism.from_sasl_name(mech.sasl_name)
+
+                # For some reason macOS sometimes returns this for mechs
+                if not (
```
frozencemetery, line 428: (This one too.)

gssapi/tests/test_high_level.py
```diff
@@ -551,6 +609,7 @@ def test_copy(self):
     # doesn't actually implement it
 
     @ktu.gssapi_extension_test('rfc6680', 'RFC 6680')
+    @ktu.krb_provider_test(['mit'], 'Heimdal does not implemented for krb5')
```
frozencemetery, line 612: Can you reword (all of) these?  "does not implemented" isn't very clear.  The resulting skip message currently is:

Your GSSAPI (provider heimdal) is known to have problems with Heimdal does not implemented for krb5.

gssapi/tests/test_raw.py
```diff
@@ -424,16 +434,21 @@ def test_store_cred_acquire_cred(self):
...
 
         store_res = gb.store_cred(deleg_creds, usage='initiate',
+                                  mech=gb.MechType.kerberos,
                                   set_default=True, overwrite=True)
         self.assertIsNotNone(store_res)
-        self.assertEqual(store_res.usage, "initiate")
-        self.assertIn(gb.MechType.kerberos, store_res.mechs)
+
+        if self.realm.provider.lower() != 'heimdal':
+            # Heimdal return this info properly
```
frozencemetery, line 442: "returns"?

gssapi/tests/test_raw.py
```diff
@@ -424,16 +434,21 @@ def test_store_cred_acquire_cred(self):
...
+            self.assertEqual(store_res.usage, "initiate")
+            self.assertIn(gb.MechType.kerberos, store_res.mechs)
 
         deleg_name = gb.inquire_cred(deleg_creds).name
         acq_resp = gb.acquire_cred(deleg_name, usage='initiate')
         self.assertIsNotNone(acq_resp)
 
     @ktu.gssapi_extension_test('cred_store', 'credentials store')
+    @ktu.krb_provider_test(['mit'], 'FIXME')
```
frozencemetery, line 451: I'd prefer not to commit "FIXME"s into the project :)

Can you adjust this to mention something about Heimdal not implementing the cred store extensions instead?

gssapi/tests/test_raw.py
```diff
@@ -629,25 +671,41 @@ def test_rfc5587(self):
...
+            test_attrs = [
+                # Taken from lib/gssapi/mech/gss_oid.c
+                [gb.OID.from_int_seq("1.3.6.1.5.5.13.24"),
+                 b"GSS_C_MA_CBINDINGS", b"channel-bindings",
+                 b""],
+                [gb.OID.from_int_seq("1.3.6.1.5.5.13.1"),
+                 b"GSS_C_MA_MECH_CONCRETE", b"concrete-mech",
+                 b"Indicates that a mech is neither a pseudo-mechanism nor a "
+                 b"composite mechanism"]
+            ]
```
frozencemetery, line 708: So these two blocks almost match.  Maybe instead of having two stanzas, you could add a `if heimdal: fix descriptions` condition?

gssapi/tests/test_raw.py
```diff
@@ -663,18 +721,30 @@ def test_sasl_names(self):
...
             self.assertIsInstance(out_mn, bytes)
 
             out_md = out.mech_description
             self.assertIsInstance(out_md, bytes)
 
-            cmp_mech = gb.inquire_mech_for_saslname(out_smn)
-            self.assertIsNotNone(cmp_mech)
-            self.assertEqual(cmp_mech, mech)
+            # Heimdal fails with Unknown mech-code on sanon
+            if not (
```
frozencemetery, line 735: (this one too)

gssapi/tests/test_raw.py
```diff
@@ -663,18 +721,30 @@ def test_sasl_names(self):
...
+                self.realm.provider.lower() == 'heimdal' and
+                mech.dotted_form == '1.3.6.1.4.1.5322.26.1.110'
+            ):
+                cmp_mech = gb.inquire_mech_for_saslname(out_smn)
+                self.assertIsNotNone(cmp_mech)
+
+                # For some reason macOS sometimes returns this for mechs
+                if not (
```
frozencemetery, line 743: (and this one)

gssapi/tests/test_raw.py
```diff
@@ -920,16 +998,21 @@ def test_krb5_ccache_name(self):
 
         old_ccache = gb.krb5_ccache_name(new_ccache.encode('utf-8'))
         try:
-            self.assertEqual(old_ccache.decode('utf-8'), self.realm.ccache)
+            if provider == 'heimdal':
+                # Heimdal never returns the old name -see above link
```
frozencemetery, line 1002: whitespace needed after hyphen

gssapi/tests/test_raw.py
```diff
@@ -1648,6 +1755,7 @@ def test_basic_iov_wrap_unwrap_autoalloc(self):
         self.assertEqual(init_message[3].value, init_other_data)
 
     @ktu.gssapi_extension_test('dce_aead', 'DCE (AEAD)')
+    @ktu.krb_provider_test(['mit'], 'unwrap fail on Heimdal')
```
frozencemetery, line 1758: Same problem with these, I think:

Your GSSAPI (provider heimdal) is known to have problems with unwrap fail on Heimdal

frozencemetery (changes requested):
There are several "if not heimdal" or "if not darwin" checks - are you intentionally assuming that macOS implies distro heimdal?  It's tricky because I think there are three different GSSAPIs one could be using on macOS.  I also don't know whether this is worth caring about.

Additional thoughts inline.

ci/run-on-linux.sh
jborean93: Yea on Heimdal it seemed to always just default to the `BOX` realm and I couldn't figure out a way to override that behaviour. Just seemed to work but making sure the hostname ended with the same realm as what we are testing with.

gssapi/tests/test_high_level.py
jborean93: Sorry force of habit is to use this style (black has trained me too well :)). I'll fix it up.

gssapi/tests/test_raw.py
jborean93: Ah this was a mistake, I was testing out an issue on macOS and forgot to remove this once I figured it out. Thanks for the pickup.

jborean93:
> are you intentionally assuming that macOS implies distro heimdal?

Somewhat, as you have said someone could have MIT Kerberos, GSS.framework Heimdal, or pure Heimdal installed on macOS so there's no easy way to detect this. I could reuse `GSSAPI_KRB5_MAIN_LIB` as the check for GSS.framework Heimdal on macOS but at this point it seemed overkill for a very niche scenario. It would also mean that tests will fail on someone who is running with GSS.framework by default as now they would need to set that env var to something that is expected. IMO I think the chances of someone running tests on macOS and they are not using GSS.framework are quite low and not worth considering until someone does encounter it.

> Additional thoughts inline.

Thanks, the changes have been made.

frozencemetery (approved):
Thanks!

Issue #259, opened 2021-08-04
python 3.10 deprecates distutils


frozencemetery:
python-3.10 implements [PEP 632](https://www.python.org/dev/peps/pep-0632/) ("Deprecate distutils module").  However, our Windows build currently depends on distutils in order to monkey-patch it: https://github.com/pythongssapi/python-gssapi/blob/main/setup.py#L79-L85

@aiudirog Do you think you'd be able to look into this?  We've got some time, but it'll presumably start throwing warnings on 3.10 release.

aiudirog:
It looks like setuptools is now vendoring `distutils` internally as `setuptools._distutils`. In theory, we should just try to apply the patch to that one and fallback to the plain `distutils` module if it isn't available.

jborean93:
Looking at https://github.com/pypa/setuptools/blob/a4dbe3457d89cf67ee3aa571fdb149e6eb544e88/docs/deprecated/distutils-legacy.rst and it looks like we can do something like

```python
os.environ['SETUPTOOLS_USE_DISTUTILS'] = 'local'

# import setuptools (which will load the shim here)
import disutils
```

This should avoid loading in the deprecated distutils in the stdlib and use the vendored copy of distutils in setuptools. Might be a good idea to see if https://github.com/pythongssapi/python-gssapi/blob/main/setup.py#L79-L85 is still required and whether we can achieve the same thing through `setuptools` directly at some point in the future.

aiudirog:
That sounds good to me. Open a PR for 3.10 support?

jborean93:
Yea now that GHA has released 3.10 I was going to modify https://github.com/pythongssapi/python-gssapi/pull/263 then look into this. Because you know more about the Windows stuff it would be great if you could validate that env var does what it should on 3.10.

jborean93:
https://github.com/pythongssapi/python-gssapi/pull/265 should solve this issue.

PR #265, opened 2021-10-05
Use distutils provided by setuptools

jborean93:
The distutils module provided by the stdlib has been deprecated in
Python 3.10. This PR sets an env var that is read by setuptools which
will shim its vendored copy of distutils if present avoiding a failure
once distutils is removed from the stdlib.

Fixes https://github.com/pythongssapi/python-gssapi/issues/259


jborean93:
Thanks for testing it out @aiudirog

Issue #230, opened 2020-11-13
MacOS wheels


frozencemetery:
### What went wrong?

For the same reason that we have Windows wheels, we should also have MacOS wheels.  Unlike Linux, we can actually do this today without too much work.

### How do we reproduce?

`pip install gssapi` on MacOS, noting the lack of wheels.

### Component versions (python-gssapi, Kerberos, OS / distro, etc.)

MacOS and probably system Heimdal.

jborean93:
[v1.7.1rc1](https://pypi.org/project/gssapi/1.7.1rc1/#description) has been uploaded and contains wheels for macOS on both `x86_64` and `arm64`. The final release of `1.7.1` will be published once the release candidate has been validated.

PR #267, opened 2021-10-06
Add macOS wheels and simplify release process

jborean93:
This changes the github workflows to an actual yaml definition (my personal preference over the quoting hell that is json).

This now uses [cibuildwheel](https://github.com/pypa/cibuildwheel) to build the Windows and macOS wheels instead of it being a manual step. The release process on PyPI is now one step instead of the manual twine call.

Fixes https://github.com/pythongssapi/python-gssapi/issues/230


jborean93:
The wheels and sdist to test from this release process. Will need to manually verify this still works on Windows and macOS before merging this PR. The artifact is from https://github.com/pythongssapi/python-gssapi/pull/267/checks?check_run_id=3811298187 which essentially ran the `release.yml` workflow just without the publishing steps.

[artifact.zip](https://github.com/pythongssapi/python-gssapi/files/7291097/artifact.zip)

jborean93:
@aiudirog if you get a chance it would be great if you could validate some of the Windows wheels in [artifact.zip](https://github.com/pythongssapi/python-gssapi/files/7291097/artifact.zip). This PR changes the build process to use [cibuildwheel](https://github.com/pypa/cibuildwheel) and while I don't forsee any problems with the changes it would be nice to have that validated by someone who uses the library.

I've done a comparison on some of the `.pyd` files in the cibuildwheel wheels and ones from the latest release on PyPI and I couldn't find any linking differences which is a good sign. I was also able to install and import them just fine with Python 3.9 but that was a very basic sanity check.

jborean93:
I've manually tested the `macOS` wheels on an M1 Mac Mini and the wheels generated install and import correctly on both the `arm64` and `x86_64` Python installs I have there. Using `otool` I can validate the dynamic links are the same with a manually built wheel from the sdist that currently occurs with `pip install gssapi` right now.

I feel confident the macOS wheels here will be fine going forward.

aiudirog:
This looks great! Can we do an RC release (either 1.8.0rc1 or 1.7.1rc1, not sure where we are semver speaking) so the wheels get built this way and uploaded to PyPI but not set as the default for the public yet? Then I can run some tests on them exactly as they will be built.

jborean93:
That's a good idea, I'll probably settle on `1.7.1rc1` as functionality there are no changes to the library, just the new wheel. There shouldn't be any breaking changes and hopefully the rc will prove that.

PR #278, opened 2022-01-11
Specify units in docstrings for all lifetime fields/parameters.

judilsteve:
A better solution would be to add properties like `lifetime_ts` that return the lifetime as a `timespan`, and maybe accept lifetime values as timestamps, with a quick `lifetime = lifetime.total_seconds if isinstance(lifetime, timespan) else lifetime` at the top of any function that accepts a lifetime parameter.

But for now this should at least avoid confusing people.

### DIFF
--- a/gssapi/creds.py
+++ b/gssapi/creds.py
@@ -74,7 +74,7 @@ class Credentials(rcreds.Creds):
 
     @property
     def lifetime(self):
-        """Get the remaining lifetime of these credentials"""
+        """Get the remaining lifetime of these credentials, in seconds"""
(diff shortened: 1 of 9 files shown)


jborean93 (approved):

jborean93:
Thanks @judilsteve for the PR, when you get a chance are you able to sign off this commit with

```bash
git commit --amend --no-edit --signoff
git push --force-with-lease origin in_seconds
```

jborean93:
These fixes were incorporated with https://github.com/pythongssapi/python-gssapi/pull/285.

PR #280, opened 2022-01-22
Allow overriding SETUPTOOLS_USE_DISTUTILS

Gu1nness:
Necessary for the package to build on debian.
Cf https://github.com/pypa/setuptools/issues/2956


jborean93:
This is not something we want to add into the `setup.py` file. It's an environment specific setting and really just a workaround for a bug that either Debian or pip need to fix on their end. In any case setuptools actually defaults to `local` by default which is where the problem stems. Using `stdlib` will only work until Python removes distutils in the stdlib which is coming in a future version.

PR #284, opened 2022-02-07
Add Python type annotation

jborean93:
Adds type annotations to the library for better type analysis and IDE integration with this library. Currently the `gssapi.raw` contents have been annotated with `.pyi` files. Using a separate file is quite useful for C extensions as an IDE can easily inspect what functions are meant to be available, their types, and doc strings without having the extension actually compiled and installed. Because of this all the docstrings have been moved out of the Cython pyx files into their respective pyi files. The Sphinx doc generator has also been updated to support this.

The next step is to annotate the high level `gssapi` package and enable some mypy tests for CI to ensure this library is valid.


simo5:
LGTM

simo5 (approved):
I think this is a mess of a change, but also very beneficial, I'll steal this stuff.

Issue #294, opened 2022-07-17
GSSAPI won't install on Python 3.11 beta


ronf:
### What went wrong?

I'm attempting to install gssapi against Python 3.11.0b3 and and getting the following errors:

```
Building wheels for collected packages: gssapi
  Building wheel for gssapi (setup.py) ... error
  ERROR: Command errored out with exit status 1:
   command: /opt/local/bin/python3.11 -u -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/tmp/pip-install-sossa1e3/gssapi_c71088ddc1c44a169aa9725a44abae04/setup.py'"'"'; __file__='"'"'/private/tmp/pip-install-sossa1e3/gssapi_c71088ddc1c44a169aa9725a44abae04/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' bdist_wheel -d /private/tmp/pip-wheel-6fnsrkam
       cwd: /private/tmp/pip-install-sossa1e3/gssapi_c71088ddc1c44a169aa9725a44abae04/
  Complete output (121 lines):
  In distributed package, building from C files...
  Skipping the s4u extension because it is not supported by your GSSAPI implementation...
  Skipping the cred_store extension because it is not supported by your GSSAPI implementation...
  Skipping the rfc4178 extension because it is not supported by your GSSAPI implementation...
  Skipping the rfc5587 extension because it is not supported by your GSSAPI implementation...
  Skipping the rfc5588 extension because it is not supported by your GSSAPI implementation...
  Skipping the cred_imp_exp extension because it is not supported by your GSSAPI implementation...
  Skipping the dce extension because it is not supported by your GSSAPI implementation...
  Skipping the dce_aead extension because it is not supported by your GSSAPI implementation...
  Skipping the iov_mic extension because it is not supported by your GSSAPI implementation...
  Skipping the ggf extension because it is not supported by your GSSAPI implementation...
  Skipping the set_cred_opt extension because it is not supported by your GSSAPI implementation...
  Skipping the rfc6680 extension because it is not supported by your GSSAPI implementation...
[...]

PR #295, opened 2022-08-03
Bump new version for 3.11 beta fixes

jborean93:
The cythonised files in the current release won't be able to compile with the latest Python 3.11 due to a changed exception struct definition. Re-cythonising with the current version produces c files that can be compiled on the latest Python version without any changes to the code being needed.

No wheels are being produced for 3.11 as it's still in beta.

Fixes https://github.com/pythongssapi/python-gssapi/issues/294


jborean93:
Going to test this on my own fork to avoid running it twice until I can get it working.

PR #296, opened 2022-08-08
Change sdist and build to use PEP517

jborean93:
Changes the sdist to no longer include the cythonised .c files and rely
on the PEP517 features to define Cython as a build requirement. This
ensures that when building this package from the sdist, the process will
include any bugfixes for Cython that have been fixed after the
python-gssapi version has been released.

The removal of the .c files also simplifies the setup.py by removing
more functionality that was tied specifically to setuptools or even
distutils making it more flexible for the future.

The CI process has been updated to test against this new artifact rather
than testing against an build_ext --inplace instance to better replicate
how end users will be using the library.

Fixes: https://github.com/pythongssapi/python-gssapi/issues/294


jborean93:
@simo5 just an FYI on this change. It's fairly large as it changes how CI works but the bulk of the stuff is moving to a PEP 517 compliant setup which allows us to simplify the sdist and wheel creation steps.

Let me know if you wish to review this or have a look through. I'm hoping to merge it in by the end of the week.

Issue #299, opened 2022-08-15
Installing v1.8.0 from sdist fails with old pip versions


webknjaz:
### What went wrong?

Whenever a pip version that's used doesn't support PEP 517/518, it just [calls `setup.py` commands](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html) instead. This means that the dependencies declared in `pyproject.toml` don't get pre-installed.

This is what's been happening in Ansible Core CI since Friday.

### How do we reproduce?

<details>
<summary><strong>Click to expand</strong> the repro log</summary>

```console
$ podman run --rm -it ubi8/ubi:8.6 sh -c 'set -x; dnf install -y python3-pip; python3 -V; pip3.6 install gssapi==1.8.0'
+ dnf install -y python3-pip
Updating Subscription Management repositories.
Unable to read consumer identity

This system is not registered with an entitlement server. You can use subscription-manager to register.

Red Hat Universal Base Image 8 (RPMs) - BaseO 2.8 MB/s | 804 kB     00:00    
Red Hat Universal Base Image 8 (RPMs) - AppSt 6.8 MB/s | 3.0 MB     00:00    
Red Hat Universal Base Image 8 (RPMs) - CodeR 130 kB/s |  20 kB     00:00    
Dependencies resolved.
==============================================================================
 Package             Arch   Version                     Repository       Size
==============================================================================
Installing:
 python3-pip         noarch 9.0.3-22.el8                ubi-8-appstream  20 k
Installing dependencies:
 platform-python-pip noarch 9.0.3-22.el8                ubi-8-baseos    1.6 M
 python3-setuptools  noarch 39.2.0-6.el8                ubi-8-baseos    163 k
 python36            x86_64 3.6.8-38.module+el8.5.0+12207+5c5719bc
                                                        ubi-8-appstream  19 k
Enabling module streams:
 python36                   3.6                                              

Transaction Summary
==============================================================================
Install  4 Packages

Total download size: 1.8 M
Installed size: 7.2 M
Downloading Packages:
(1/4): python3-pip-9.0.3-22.el8.noarch.rpm    230 kB/s |  20 kB     00:00    
(2/4): python36-3.6.8-38.module+el8.5.0+12207 465 kB/s |  19 kB     00:00    
(3/4): python3-setuptools-39.2.0-6.el8.noarch 1.1 MB/s | 163 kB     00:00    
[...]

PR #300, opened 2022-08-15
Bump version to 1.8.1 with Python 3.7 requires

jborean93:
This update bumps the `python_requires` to be `>=3.7` as the sdist build
requires PEP 517 support that is somewhat only guaranteed with the pip
that was shipped with Python 3.7 or greater. Since 3.6 has been end of
life since December 2021 and the remaining Linux distributions that
still use it have system provided packages this shouldn't be a surprise.

Fixes https://github.com/pythongssapi/python-gssapi/issues/299

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,8 +57,6 @@ jobs:
           version: cp38-macosx_arm64
         - os: macOS-12
           version: cp37-macosx_x86_64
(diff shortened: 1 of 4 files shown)


webknjaz (commented):
Don't forget to drop the trove classifier too.

jborean93:
Thanks, I forgot about those, they've been dropped.

Issue #212, opened 2020-06-30
Building 'gssapi.raw.ext_cred_store' extension on SLES 15.1 stops with "gssapi/raw/ext_cred_store.c:1333:8: error: unknown type name ‘gss_key_value_set_desc’"


RolandFelnhofer:
### What went wrong?
When building python-gssapi ist stops with "gssapi/raw/ext_cred_store.c:1333:8: error: unknown type name ‘gss_key_value_set_desc’"

### How do we reproduce?
```
$ cd /usr/src/gssapi-1.6.5
$ python3 -mvenv buildenv
$ source buildenv/bin/activate
$ mkdir wheels
$ export PATH=$PATH:$(dirname $(locate -r "krb5-config$"))
$ pip install -U pip wheel setuptools
$ pip wheel -w wheels/ pip wheel setuptools
$ pip wheel -w wheels/ pip wheel gssapi
```

### Component versions (python-gssapi, Kerberos, OS / distro, etc.)
python-gssapi: 1.6.5
OS: SUSE Linux Enterprise Server 15 SP1
```
$ krb5-config --all
Version:     Kerberos 5 release 1.16.3
Vendor:      Massachusetts Institute of Technology
Prefix:      /usr/lib/mit
Exec_prefix: /usr/lib/mit
```

### Reason
I found out that in setup.py, between line 103 and 114, tries to detect the location of 'gssapi_ext.h', by executing 'krb5-config gssapi --prefix' and adding 'include/gssapi/gssapi_ext.h' to that prefix. 

On SLES 15.1 (and other SUSE distributions I assume) the location of the include files for gssapi are within the regular include path ('/usr/include'). So the full path for the header file is '/usr/include/gssapi/gssapi_ext.h' and not '/usr/lib/mit/include/gssapi/gssapi_ext.h' as expected by 'setup.py'

Can you please add a check if 'gssapi_ext.h' exists within the regular include structure?

### Workaround
As a temporary workaround I used:
```
$ pip wheel --global-option=build_ext --global-option="-DHAS_GSSAPI_EXT_H" -w wheels/ gssapi
```
to force to include the header file 'gssapi_ext.h'

frozencemetery:
I'm pretty sure this is a bug in SUSE.  This should match to the autoconf prefix from building krb5 - see https://github.com/krb5/krb5/blob/master/src/build-tools/krb5-config.in#L30

For completeness, here are some other distros and their values:

Debian testing:
```
[...]

PR #320, opened 2023-06-17
Try finding gssapi_ext.h includedir via pkg-config

mweinelt:
Gives distros a chance, that put the development headers in a path, that krb5-config does not expose, but for which pkg-config could possibly emit an include directory path.

Related to issues like https://github.com/pythongssapi/python-gssapi/issues/212 and follows up on the train of thought in https://github.com/pythongssapi/python-gssapi/issues/212#issuecomment-941902841.

I don't expect any behavior changes, since we still consider `krb5-config` the primary source of information, but if it doesn't find anything we try the includedir `pkg-config` provides.

I have confirmed this fixes the build on nixpkgs after providing pkg-config.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -124,17 +124,26 @@ if compile_args is None:
 
 # add in the extra workarounds for different include structures
 if winkrb_path:
-    prefix = winkrb_path
+    prefixes = [winkrb_path]
 else:
+    prefixes = []
     try:
-        prefix = get_output(f"{kc} gssapi --prefix")
+        prefixes.append(get_output(f"{kc} gssapi --prefix"))
     except Exception:
-        print("WARNING: couldn't find krb5-config; assuming prefix of %s"
-              % str(sys.prefix))
-        prefix = sys.prefix
-gssapi_ext_h = os.path.join(prefix, 'include/gssapi/gssapi_ext.h')
-if os.path.exists(gssapi_ext_h):
-    compile_args.append("-DHAS_GSSAPI_EXT_H")
+        print("WARNING: couldn't find krb5-config to determine prefix")
+    try:
+        prefixes.append(
+            get_output("pkg-config --variable=includedir krb5-gssapi")
+        )
+    except subprocess.CalledProcessError:
+        print("WARNING: unable to find includedir using pkg-config")
+    prefixes.append(sys.prefix)
+
+for prefix in prefixes:
+    gssapi_ext_h = os.path.join(prefix, 'gssapi/gssapi_ext.h')
+    if os.path.exists(gssapi_ext_h):
+        compile_args.append("-DHAS_GSSAPI_EXT_H")
+        break
 
 # Create a define to detect msys in the headers
 if sys.platform == 'msys':


setup.py
```diff
@@ -124,17 +124,26 @@ def get_output(*args, **kwargs):
...
+    try:
+        prefixes.append(os.path.dirname(
+            get_output("pkg-config --variable=includedir krb5-gssapi"))
+        )
+    except Exception:
+        print("WARNING: could not find pkg-config to determine includedir")
+    prefixes.append(sys.prefix)
+
+for prefix in prefixes:
+    gssapi_ext_h = os.path.join(prefix, 'include/gssapi/gssapi_ext.h')
```
jborean93, line 143: On Fedora 38 the `pkg-config` output is `/usr/include` so the check would ultimately be for `/usr/include/include/gssapi/gssapi_ext.h`. It will still work because as you said it will check the `krb5-config gssapi --prefix` first but it sounds like the logic here might need some extra work.

Even testing on SLES 15 shows that it returns `/usr/include` so this fix won't work on the platforms that might benefit from it a bit more.

setup.py
```diff
@@ -124,17 +124,26 @@ def get_output(*args, **kwargs):
...
-        print("WARNING: couldn't find krb5-config; assuming prefix of %s"
-              % str(sys.prefix))
-        prefix = sys.prefix
-gssapi_ext_h = os.path.join(prefix, 'include/gssapi/gssapi_ext.h')
-if os.path.exists(gssapi_ext_h):
-    compile_args.append("-DHAS_GSSAPI_EXT_H")
+        print("WARNING: couldn't find krb5-config to determine prefix")
+    try:
+        prefixes.append(os.path.dirname(
+            get_output("pkg-config --variable=includedir krb5-gssapi"))
```
jborean93, line 136: I don't know too much about `pkg-config` but is `krb5-gssapi` unique enough across the platforms? What governs this id and are we always sure it's going to match krb5/gssapi. It also brings into question how does it deal with Heimdal.

It should also probably deal with an empty string being outputted as well.

jborean93 (commented):
I'm sorry this PR slipped my notice so wasn't able to review it.

I've left a few comments but I'm still concerned this could break scenarios where both MIT krb5 and Heimdal are installed. In this case MIT krb5 includes `gssapi_ext.h` but Heimdal does not. If Heimdal was selected to link against then it might be compiled with `-DHAS_GSSAPI_EXT_H"` because the MIT krb5 `pkg-config` entry is found.

It is a rare scenario but still something that needs to be taken into consideration.

setup.py
mweinelt: It works with both packages on NixOS, and if it is missing it returns and empty string to stdout. I don't have any other machines at this time I could test on.

```
$ nix-shell -p heimdal.dev pkg-config --run "pkg-config --variable=includedir krb5-gssapi"
/nix/store/6w2xlb77p6y8p3amb7q1kw3pbdm4kkmx-heimdal-7.8.0-dev/include

$ nix-shell -p krb5.dev pkg-config --run "pkg-config --variable=includedir krb5-gssapi"
/nix/store/hwv4s9dxmwg7ahzjzlarhwjkirvwzlyb-krb5-1.20.1-dev/include

$ nix-shell -p pkg-config --run "pkg-config --variable=includedir krb5-gssapi"
Package krb5-gssapi was not found in the pkg-config search path.
Perhaps you should add the directory containing `krb5-gssapi.pc'
to the PKG_CONFIG_PATH environment variable
No package 'krb5-gssapi' found
```

The name is governed by the pkg-config file

```
$ cat /nix/store/lg22dp1sij739swldvrf7410r2j209l0-heimdal-7.8.0-dev/lib/pkgconfig/krb5-gssapi.pc
prefix=/nix/store/cy2j8bfp8dckdxaf2k9m0d3k4mqrpgrl-heimdal-7.8.0
exec_prefix=/nix/store/cy2j8bfp8dckdxaf2k9m0d3k4mqrpgrl-heimdal-7.8.0
libdir=/nix/store/cy2j8bfp8dckdxaf2k9m0d3k4mqrpgrl-heimdal-7.8.0/lib
includedir=/nix/store/lg22dp1sij739swldvrf7410r2j209l0-heimdal-7.8.0-dev/include
vendor=Heimdal

Name: krb5-gssapi
Description: Kerberos implementation of the GSS API.
Version: 7.8.0
Requires: heimdal-gssapi

$ cat /nix/store/y9rk7j5afwwzsjk6ximhipngbap8v8vg-krb5-1.20.1-dev/lib/pkgconfig/krb5-gssapi.pc
prefix=/nix/store/nnkw7wdy3f5nlw912hksc82n15s7cwkd-krb5-1.20.1
exec_prefix=${prefix}
libdir=/nix/store/nnkw7wdy3f5nlw912hksc82n15s7cwkd-krb5-1.20.1/lib
includedir=/nix/store/y9rk7j5afwwzsjk6ximhipngbap8v8vg-krb5-1.20.1-dev/include
vendor=MIT

Name: krb5-gssapi
Description: Kerberos implementation of the GSSAPI
Version: 1.20.1
Requires: mit-krb5-gssapi
```

mweinelt: > It should also probably deal with an empty string being outputted as well.

In case pkg-config does not find anything, it exits with 1. I can catch `CalledProcessError` and that should be fine, no?

jborean93: > The name is governed by the pkg-config file

But what is setting that name, is it the package maintainers, do they all state that this package should be `krb5-gssapi` or are there distros where that is not the case.

> In case pkg-config is unavailable or does not find anything, it exits with 1. I can catch CalledProcessError and that should be fine, no?

I'm talking about if `pkg-config` is available but the configure value is an empty string. I have no idea if this is a valid scenario but by adding an empty string the header check will check in `include/gssapi/...` relative to the current dir rather than it being rooted somewhere.

setup.py
mweinelt: The call to `os.path.dirname` is what was wrong, will push a fixed version in a few minutes.

mweinelt:
> this could break scenarios where both MIT krb5 and Heimdal are installed

I'm not sure how to tackle this.

jborean93:
> I'm not sure how to tackle this.

That's ultimately why the current approach is to use the values from `krb5-config` in the path. This should give us the prefix for the found `krb5-config`. It's just that it seems like there was a bug on SLES where it was returning `/usr/lib/mit` which was how SLES configured the project before moving the files. I think greghudson covers this pretty succinctly with https://github.com/pythongssapi/python-gssapi/issues/212#issuecomment-652452507 in that there are multiple things that might be doing things sub optimally.

Ultimately when testing an install with the SLES 15.1 container image, the krb5 package is now reporting the prefix as `/usr` like all the other distros. I'm unsure when the fix was introduced but it certainly is there. Based on the current logic I'm not sure it is going to really be possible to have fallback locations. As a workaround for older Suse versions that might still be affected then I think maybe adding an env var override like `GSSAPI_EXT_H_PATH` that allows a user to specify a custom location to check for the `gssapi_ext.h` header might be a better approach. Essentially things should work on newer versions but we have a workaround where that might not be the case.

setup.py
mweinelt: > But what is setting that name, is it the package maintainers, do they all state that this package should be krb5-gssapi or are there distros where that is not the case.

The names are upstream-maintained.

https://github.com/heimdal/heimdal/blob/f62e2f278437ff6c03d2d09bd628381c795bba78/tools/krb5-gssapi.pc.in#L7
https://github.com/krb5/krb5/blob/0f870b1bcad960fd5319a3f97aafd7f4a289e2fb/src/build-tools/krb5-gssapi.pc.in#L7

> I'm talking about if pkg-config is available but the configure value is an empty string. I have no idea if this is a valid scenario but by adding an empty string the header check will check in include/gssapi/... relative to the current dir rather than it being rooted somewhere.

Given that these files are maintained by upstream, the risk of running into empty values should be negligible.

PR #331, opened 2023-10-30
Update setup.py

sanglist:
Small fix to enable vas_gssapi
Signed-off-by: Shlomo Anglister  <shlomo.anglister@intel.com>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -181,7 +181,7 @@ if ENABLE_SUPPORT_DETECTION:
             os.environ['PATH'] += os.pathsep + os.path.dirname(main_lib)
     elif main_lib is None:
         for opt in libraries:
-            if opt.startswith('gssapi'):
+            if 'gssapi' in opt:
                 if os.name == 'nt':
                     main_lib = '%s.dll' % opt
                     if winkrb_path:


sanglist:
Please merge to resolve setup issue

simo5:
@jborean93 this patch looks sane enough, ok if I merge it?

jborean93:
Sorry this one was lost in my sea of notifications. At a glance it seems ok but I do wonder if it might potentially pick up some false positives.

jborean93:
I only want to approve it when I know it won't break other use cases. What setup is causing the problem here, what is the output of `krb5-config --libs gssapi` and is there potentially a better way of fixing your problem.

sanglist:
> I only want to approve it when I know it won't break other use cases. What setup is causing the problem here, what is the output of `krb5-config --libs gssapi` and is there potentially a better way of fixing your problem.

On SLES12 (SuSE 12 Enterprise)

-Wl,-rpath,/opt/quest/lib64:/opt/quest/lib64/support -L/opt/quest/lib64 -lvas -L/opt/quest/lib64/support -lgcc_s -lvas-gssapi -lrt -lresolv

jborean93:
Thanks, is `vas-gssapi` just a SLES thing or is it a specific package here? I'm afraid I'm not familiar with it.

sanglist:
It depends on corporate, right ?
bridging all kinds of authentication protocols originating from a vast application space to corporate authentication.

jborean93:
What I’m trying to get at is how can I test out this scenario. If I need to maintain this I need to know who could be affected and what scenarios this code needs to handle. Do you have any info on this vas-gssapi library. Is it a product I can install or is it a paid product. Are there docs online about it that you can share.

sanglist:
The solution at: [pykrb5/setup.py](https://github.com/jborean93/pykrb5/blob/main/setup.py)
Seems to provide a one size fits all using environment variables.

PR #338, opened 2024-01-28
Enable the use of `mypy --no-explicit-reexport` on downstream projects

dkg:
When testing simple python code like this with mypy for type-safety:

```
import gssapi
gsc:gssapi.SecurityContext
gsc = gssapi.SecurityContext(
       usage='initiate',
       name=gssapi.Name('imap@localhost',
                        gssapi.NameType.hostbased_service))
```

I see these errors:

```
0 $ mypy --no-implicit-reexport ./test.py
test.py:3: error: Name "gssapi.SecurityContext" is not defined  [name-defined]
test.py:4: error: Module "gssapi" does not explicitly export attribute "SecurityContext"  [attr-defined]
test.py:4: error: Module "gssapi" does not explicitly export attribute "Name"  [attr-defined]
test.py:4: error: Module "gssapi" does not explicitly export attribute "NameType"  [attr-defined]
Found 4 errors in 1 file (checked 1 source file)
1 $
```

The same thing happens when using `mypy --strict`.

[a blogpost](https://til.codeinthehole.com/posts/how-to-handle-convenience-imports-with-mypy/) suggested that the `__all__` variable in `gssapi/__init__.py` might be the way to fix this.

I can confirm that it does clear the error, but I'm not enough of a python module expert to know whether there might be some undesirable side effects as well to making this change.  Please review!

Tested on debian testing/unstable, with:

- python3  3.11.6-1
- python3-gssapi 1.8.2-1+b2
- mypy 1.8.0-1

### DIFF
--- a/gssapi/__init__.py
+++ b/gssapi/__init__.py
@@ -38,3 +38,17 @@ from gssapi.sec_contexts import SecurityContext  # noqa
 from gssapi.mechs import Mechanism  # noqa
 
 from gssapi._utils import set_encoding  # noqa
+
+__all__ = [
+    'AddressType',
+    'Credentials',
+    'IntEnumFlagSet',
+    'Mechanism',
+    'MechType',
+    'Name',
+    'NameType',
+    'OID',
+    'RequirementFlag',
+    'SecurityContext',
+    'set_encoding',
+]


jborean93 (approved):
Thanks for the fix.

jborean93 (changes requested):
Thanks for the fix, looks like the line length is too long and has hit the flake sanity test. You can do something like

```python
__all__ = [
    "SecurityContext",
    "Name",
    ...
    "set_encoding",
]
```

jborean93:
I'll have to look into why the build is failing, maybe cibuildwheel pushed an update which broke something.

jborean93:
Looks like a temporary problem, rerunning a few hours later and it now works.

jborean93 (approved):

PR #341, opened 2024-01-29
Update GHA action versions

jborean93:
Updates the GitHub Actions versions to the latest ones available.


jborean93:
Will need to wait until https://github.com/pypa/cibuildwheel/issues/1740 is resolved as the Windows wheel failures are only going to get worse as 7.4 is rolled out to more runs.

Issue #347, opened 2024-08-25
cred_store extension not supported on macOS


ronf:
### What went wrong?

I'm trying to use python-gssapi on macOS 14.6 and trying to take advantage of the cred store extension. However, when I build python-gssapi, I get back:

Skipping the cred_store extension because it is not supported by your GSSAPI implementation...

### How do we reproduce?

Create a gssapi.Credentials object with the 'store' argument, such as:

```python
creds = gssapi.Credentials(usage="initiate", store={"ccache": "MEMORY:username"})
```

*(Remember to use fenced code blocks and consider placing in a gist if large)*

### Component versions (python-gssapi, Kerberos, OS / distro, etc.)

MIT Kerberos 5 version 1.21.3, from MacPorts 2.10.1
python-gssapi 1.8.3 from PyPI running on Python 3.12.5

I also tried getting the latest python-gssapi from Git (1.8.4) and installing that, but I ran into the same result.

The problem seems to be in the "support detection" in setup.py, but from what I can tell the version of Kerberos I have installed does have the symbol (gss_store_cred_into) that setup.py is looking for:

```
nm -gU /opt/local/lib/libgssapi_krb5.dylib | grep store_cred
0000000000011594 T _gss_store_cred
00000000000115c0 T _gss_store_cred_into
```

I think it might be finding the wrong library -- when I printed what it found, it reported:

/System/Library/Frameworks/GSS.framework/GSS

However, on my system that's a broken symlink to /System/Library/Frameworks/GSS.framework/Versions/Current/GSS, which points at /System/Library/Frameworks/GSS.framework/Versions/A/GSS, which doesn't seem to exist:

```
ls -l /System/Library/Frameworks/GSS.framework/Versions/A/
total 0
drwxr-xr-x  7 root  wheel  224 Aug  4 03:31 Resources/
drwxr-xr-x  3 root  wheel   96 Aug  4 03:31 _CodeSignature/
```

I then tried setting GSSAPI_MAIN_LIB=/opt/local/lib/libgssapi_krb5.dylib and running "setup.py build" rather than building with pip, but for this to work I needed to comment out a bit of setup.py:

```patch
diff --git a/setup.py b/setup.py
index a71967f..39779df 100755
--- a/setup.py
+++ b/setup.py
@@ -45,9 +45,9 @@ link_args, compile_args = [
 ]

 osx_has_gss_framework = False
-if sys.platform == 'darwin':
[...]

PR #348, opened 2024-08-26
Fix DCE symbol name logic on macOS

jborean93:
Only use the private Apple symbol name for the DCE extension detection when using the actual GSS.Framework target. This will allow compiling this library against a custom MIT krb5 or Heimdal library on macOS with support for the normal IOV wrapping functions.

Fixes: https://github.com/pythongssapi/python-gssapi/issues/347


simo5 (approved):
LGTM

jborean93:
Thanks for looking over this @simo5

PR #351, opened 2024-10-03
really drop python<=3.7 support

kloczek:
Filter all code over `pyupgracde --py38-plus`.

### DIFF
--- a/docs/custom_extensions/requires_rfc.py
+++ b/docs/custom_extensions/requires_rfc.py
@@ -8,7 +8,7 @@ def setup(app):
     app.add_role('requires-ext', RequiresExtRole(app))
 
 
-class RequiresExtRole(object):
+class RequiresExtRole:
     def __init__(self, app):
         self.app = app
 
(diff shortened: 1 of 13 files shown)


kloczek:
BTW end of this month 3.8 will be EOSed.

kloczek:
Any update?  🤔

jborean93:
Sorry I didn't get to this earlier. I'm reluctant to merging in stylistic changes without having an actual check in CI to back it for future commits. Granted there are still some Python 2-isms still in there that this PR is fixing but making these changes doesn't really achieve to much in the long run.

If I was really wanting to do something like this I would prefer to just have something like `black` to go over the codebase.

kloczek:
`black` is the Python code formatter.
`pyupgrade` replaces syntax to more effective handled by exact Python version (please read `pyupgrade` documentation for details).
`gssapi` passes pytest test suite with this PR.
Did you test this PR? 🤔

PR #352, opened 2025-02-17
Fix return type for name property in Credentials class.

DrFobos:
It seems like type annotation for [name](https://github.com/pythongssapi/python-gssapi/blob/main/gssapi/creds.py#L85) property in [gssapi.creds.Credentials](https://github.com/pythongssapi/python-gssapi/blob/main/gssapi/creds.py#L17) class should be gssapi.names.Name rather than gssapi.raw.names.Name.

Because it explicitly casts in [gssapi.creds.Credentials.inquire](https://github.com/pythongssapi/python-gssapi/blob/main/gssapi/creds.py#L308).

Change return type and add checks into tests.

### DIFF
--- a/gssapi/creds.py
+++ b/gssapi/creds.py
@@ -82,9 +82,9 @@ class Credentials(rcreds.Creds):
                       super(Credentials, cls).__new__(cls, base_creds))
 
     @property
-    def name(self) -> rnames.Name:
+    def name(self) -> names.Name:
         """Get the name associated with these credentials"""
-        return t.cast(rnames.Name,
+        return t.cast(names.Name,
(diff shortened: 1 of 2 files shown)


jborean93 (approved):

jborean93:
Hey sorry the CI workers were a bit old so I had to update them in https://github.com/pythongssapi/python-gssapi/pull/353 to get CI passing. Would you be able to rebase this PR so it runs with the latest changes before I merge it in?

DrFobos:
Sure, no problem.

jborean93 (approved):
Thanks for fixing this.

PR #358, opened 2025-04-18
Add CppProperties.json for x86-Debug configuration with IntelliSense support

Nicklihou:


jborean93:
Thanks for the contribution but we cannot accept changes with binaries and user specific paths. This might be better off being put inside some document/github gist and sharing the steps to create these files for a user's environment.

PR #362, opened 2025-06-16
remove deprecated license classifier

### LINKED ISSUE #361: warning during build because of deprecated license classifier
Trying to install (and hence build) gssapi on WSL2 (ubuntu 24.04) and getting the following warning

```
      /home/username/.cache/uv/builds-v0/.tmpE1Yu3q/lib/python3.13/site-packages/setuptools/dist.py:759:
      SetuptoolsDeprecationWarning: License classifiers are deprecated.
      !!

              ********************************************************************************
              Please consider removing the following classifiers in favor of a SPDX license expression:

              License :: OSI Approved :: ISC License (ISCL)

              See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
              ********************************************************************************

      !!
        self._finalize_license_expression()
```
which seems to correspond to [this line in `setup.py`](https://github.com/pythongssapi/python-gssapi/blob/main/setup.py#L306). Maybe this can be removed because it seems the license is also separately specified

theavey:
Should fix #361

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -303,7 +303,6 @@ setup(
         'Programming Language :: Python :: 3.12',
         'Programming Language :: Python :: 3.13',
         'Intended Audience :: Developers',
-        'License :: OSI Approved :: ISC License (ISCL)',
         'Programming Language :: Python :: Implementation :: CPython',
         'Programming Language :: Cython',
         'Topic :: Security',


jborean93:
Thanks for fixing this.

PR #370, opened 2025-12-08
Add Free-Threading and Limited API/Stable ABI

jborean93:
Add support for building wheels with the free-threaded interpreter and also with the limited API/stable ABI on Python 3.11+. There is limited testing for the free-threading setup but this become usable for people running on that interpreter without reverting back to locking the GIL.


jborean93:
We cannot bump cibuildwheel to a newer version until https://github.com/pypa/cibuildwheel/issues/2663 is fixed. Luckily `3.2.1` includes what we need right now.

PR #374, opened 2026-06-18
Add gss_localname and friends

### LINKED ISSUE #49: Implement gss_localname and friends
There are four related methods that are part of the Solaris extensions:

- gss_localname
- gss_userok
- gss_pname_to_uid
- gss_authorize_localname

(we should probably also have, optionally,  GSS_C_ATTR_LOCAL_LOGIN_USER as well)

JacobHenner:
* gss_authorize_localname
* gss_localname
* gss_pname_to_uid
* gss_userok

Fixes #49


gssapi/raw/ext_localname.pyi
```diff
@@ -0,0 +1,93 @@
+import typing as t
+
+if t.TYPE_CHECKING:
```
jborean93, line 3: Do we need this check, I don't think the names and oids files will need to import this ext file and we shouldn't need to use the `t.TYPE_CHECKING` conditional import.

gssapi/raw/ext_localname.pyx
```diff
@@ -0,0 +1,156 @@
...
+        name (Name): the GSSAPI name to check
+        username (Union[bytes, str]): the local username to check against
+
+    Returns:
+        bool: whether or not the name is authorized to act as the user
+    """
+    cdef int res
+
+    if isinstance(username, str):
+        username = username.encode(_utils._get_encoding())
```
jborean93, line 78: Is there a benefit to accepting both bytes and str here? I would have thought we would just use bytes but it's been a while since I've looked at what some of the other functions are doing.

gssapi/raw/ext_localname.pyi
JacobHenner: I was following the pattern I observed in the other ext_*.pyi files within the same directory, but perhaps I've overlooked some other aspect that establishes the necessity for this block. What's the criteria for including/not including?

jborean93: I'm fairly certain it's only there when there are circular imports, e.g. `names` or `oids` import this file. The `__init__.py` is somewhat different enough that maybe it is required here but I wouldn't have thought so.

